PR #1097, opened 2022-02-08
[Editor Permissions] Allow Editors to Censor Pdfs

calvinhlee23:
### Notes: depends on https://github.com/ResearchHub/researchhub-backend/pull/967
- See comment


config/api.js
```diff
@@ -354,7 +354,9 @@ const routes = (BASE_URL) => {
 
       return url;
     },
-
+    PAPER_CENSOR: ({ paperId, isRemoved }) =>
+      BASE_URL +
+      `paper/${paperId}/${isRemoved ? "censor_paper" : "restore_paper"}/`,
```
yattias, line 359: To follow convention, could we have the endpoint be `/censor` and `/restore` ? The paper prefix implies that a paper is the subject.

calvinhlee23: These endpoints were already being used.

lightninglu10 (approved):

PR #1104, opened 2022-02-22
Referral program

lightninglu10:
Referral program only shows for users in the top 1% of RSC (currently at 138 RSC)

### DIFF
--- a/components/Navbar.js
+++ b/components/Navbar.js
@@ -281,10 +281,7 @@ const Navbar = (props) => {
 
   useEffect(() => {
     function fetchReferrals() {
-      return fetch(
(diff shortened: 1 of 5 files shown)


components/Navbar.js
```diff
@@ -277,6 +279,24 @@ const Navbar = (props) => {
     setSideMenu(!sideMenu);
   }
 
+  useEffect(() => {
+    function fetchReferrals() {
+      return fetch(
+        api.SHOW_REFERRALS({ reputation: user.reputation }),
```
yattias, line 285: It doesn't seem like reputation needs to be passed in as an argument

yattias (approved):

PR #1113, opened 2022-03-03
[Feed][Bugfix] Prefetch Messing up the displayed Feed on quick filter updates

calvinhlee23:
- Prefetch Messing up the displayed Feed on quick filter updates
- Added memoized variable to check if it's ok to prefetch


yattias (approved):
LGTM

PR #1115, opened 2022-03-04
Redesigned feed cards to be more compact

thomas-vu:
**What?**
- Consolidated `PaperEntryCard`, `UserPostCard`, and `HypothesisCard` into `FeedCard`.
- Redesigned FeedCard to be more vertically compact.

# Before
Screen Shot 2022-03-03 at 5 57 27 PM

# After
Screen Shot 2022-03-03 at 5 54 59 PM


components/Author/Tabs/FeedCard.tsx
```diff
@@ -0,0 +1,558 @@
...
+  /**
+   * Whether or not THIS PaperPDFModal is open.
+   * There may be many PaperPDFModal components on the page, but
+   * modals.openPaperPDFModal is only a single boolean. So all cards
+   * must only render their PaperPDFModal component if requested */
+  const [isPreviewing, setIsPreviewing] = useState(false);
+  const [voteState, setVoteState] = useState<string | null>(
+    userVoteToConstant(userVote)
+  );
+  const [score, setScore] = useState<number>(initialScore + (boostAmount || 0));
```
yattias, line 108: @thomas-vu we are moving away from adding boost amount in the score because it is confusing to the user and also to us when we debug. I made that change on the post page but didn't realize we are also doing this on cards - can you remove the boost from the score?

Generally speaking, we should decouple the score from the twitter score and any other factors such as this.

components/Author/Tabs/FeedCard.tsx
```diff
@@ -0,0 +1,558 @@
...
+    return arr.filter((el) => {
+      return !isNullOrUndefined(el);
+    });
+  }
+
+  const createVoteHandler = (voteType) => {
+    const voteStrategies = {
+      [UPVOTE]: {
+        increment: 1,
+        getUrl:
```
yattias, line 131: Maybe I'm missing something but this seem to only imply a voting strategy on posts and hypotheses but what about papers?

thomas-vu: Papers have a separate callback called `onPaperVote`, this was legacy logic from PaperEntryCard I wasn't confident in changing.

components/Author/Tabs/FeedCard.tsx
thomas-vu: Sure thing, I'll remove this.

yattias (approved):

PR #1117, opened 2022-03-04
Default author when publishing and New badge for ELN features

thomas-vu:
**What?**
- When opening the publishing modal, the current user is assigned as the default author.
- Added New badge for ELN features

Screen Shot 2022-03-04 at 12 31 15 PM

Screen Shot 2022-03-04 at 12 31 10 PM


lightninglu10 (approved):

lightninglu10:
Let's remember to remove this new feature in like 2 weeks.

lightninglu10:
It would also be cool if we could add a notification dot on the profile dropdown indicating there's something new. 

May be too complex to do it well though (but if we utilize localStorage, maybe?)

PR #1118, opened 2022-03-04
[Feed][Bugfix] - Prefetch local var resetting on component rerender

calvinhlee23:
- Component re-render was resetting locally memoized variable for the prefetch function
- Made the function into a state of the top-level component
- Added force reset on the refetch on doc-type changes


yattias (approved):
LGTM

PR #1119, opened 2022-03-04
Updating start_date on feed to be based on absolute time rather than relative

yattias:
### What?
- At the moment, our time filters are off. For example, if today's date is 03-01 then "This month" will only show things that happened today. 
- I confirmed that reddit works differently. Selecting any time filter will subtract an absolute number of days from today. For example, When applying "This month" in reddit, reddit will show posts in the past 30 days.
- Doing things this way gets around the problem of different client timezones. In NYC "today" means a different epoch than in the west coast.


calvinhlee23:
This works. Maybe we should consider setting FE enums & make backend handle any timeframe related filters in the future

calvinhlee23 (approved):

PR #1123, opened 2022-03-10
Tags: UI + Search capability

zainlks:
- Tags in the UI feed
- Ability to add tags via edit screen
- Search for tags within edit screen

### DIFF
--- a/components/Author/Tabs/FeedCard.tsx
+++ b/components/Author/Tabs/FeedCard.tsx
@@ -289,11 +289,9 @@ function FeedCard(props: FeedCardProps) {
               <div className={css(styles.tagContainer)}>
                 {tags.map((tag, index) => (
                   <div className={css(styles.tag, styles.firstTag)} key={index}>
-                    <Link href={`/search/all?q=${tag.key}`}>
-                      <a className={css(styles.tagLink)}>
-                        {`#${tag.key}`}
-                      </a>
-                    </Link>
+                    <a className={css(styles.tagLink)} href={`/?tag=${tag.id}&tag_key=${tag.key}`}>
+                      {`#${tag.key}`}
+                    </a>
                   </div>
                 ))}
               </div>
--- a/components/Author/Tabs/FeedCard.tsx
+++ b/components/Author/Tabs/FeedCard.tsx
@@ -289,9 +289,9 @@ function FeedCard(props: FeedCardProps) {
               <div className={css(styles.tagContainer)}>
                 {tags.map((tag, index) => (
                   <div className={css(styles.tag, styles.firstTag)} key={index}>
-                    <Link href={`/search/all?q=${tag}`}>
+                    <Link href={`/search/all?q=${tag.key}`}>
                       <a className={css(styles.tagLink)}>
-                        {`#${tag}`}
+                        {`#${tag.key}`}
                       </a>
                     </Link>
                   </div>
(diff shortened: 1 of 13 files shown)


components/Author/Tabs/FeedCard.tsx
```diff
@@ -284,6 +286,17 @@ function FeedCard(props: FeedCardProps) {
               <span className={css(styles.title)}>
                 {titleAsHtml ? titleAsHtml : title ? title : ""}
               </span>
+              <div className={css(styles.tagContainer)}>
+                {tags.map((tag, index) => (
+                  <div className={css(styles.tag, styles.firstTag)} key={index}>
+                    <Link href={`/search/all?q=${tag}`}>
```
yattias, line 292: Would clicking the tag take the user to the search page or the feed page?

components/Author/Tabs/FeedCard.tsx
```diff
@@ -541,6 +558,26 @@ const styles = StyleSheet.create({
     fontSize: 14,
     marginRight: 5,
   },
+  tag: {
+    display: "inline-block",
+    backgroundColor: "#E8EEFD",
```
yattias, line 563: We try to use only colors defined in `colors.js` btw. I know you are waiting on design for Tetiana so this is just a heads up. Sometimes she uses colors that are slightly off our codebase in which case we figure out whether to use our own or introduce a new color in the palette.

components/Author/Tabs/FeedCard.tsx
```diff
@@ -541,6 +558,26 @@ const styles = StyleSheet.create({
...
+    color: "#4F53F6",
+    fontWeight: 600,  
+    padding: "6px 20px",
+    borderRadius: "18px",
+    margin: "0px 10px 10px 0px",
```
yattias, line 568: Maybe you already given it some thought and this is already addressed here. Wondering if there will be a bunch of empty space in the case a feed card has no tags

components/Paper/Upload/api/useEffectGetSuggestedTags.ts
```diff
@@ -0,0 +1,28 @@
+import {
+  emptyFncWithMsg,
+  isNullOrUndefined,
+  nullthrows,
+} from "../../../../config/utils/nullchecks";
```
yattias, line 5: BTW, we implemented some config change recently that allows you to use "~" to refer to the root dir. You can do `~/config/utils/nullchecks` instead

components/Paper/Upload/api/useEffectGetSuggestedTags.ts
```diff
@@ -0,0 +1,28 @@
...
+import { getSuggestedTags } from "./getSuggestedTags";
+import { useEffect } from "react";
+
+type Payload = {
+  momoizeOn?: any[];
```
yattias, line 10: I don't see `memoizeOn` being passed as an arg

components/Paper/Upload/api/getSuggestedTags.ts
```diff
@@ -0,0 +1,28 @@
...
+  searchString: string;
+};
+
+export function getSuggestedTags({ onError, onSuccess, searchString }: Payload): void {
+  fetch(API.TAG({ limit: 1000, search: searchString }), API.GET_CONFIG())
+    .then(Helpers.checkStatus)
+    .then(Helpers.parseJSON)
+    .then((resp: any): void => {
+      const tags = resp.tags
+        .map((tag, _index) => {
```
calvinhlee23, line 16: I know you're following the pattern of getSuggestedHubs here which I happened to write based on some legacy code for convenience. :/.
I was wondering if we can migrate this to BE & we debounce the request instead of fetching all of the tags all at once.

PR #1127, opened 2022-03-18
Added Crossref DOIs to publishing (Frontend)

thomas-vu:
Screen Shot 2022-03-23 at 6 58 49 PM

### DIFF
--- a/components/Modals/NotePublishModal.tsx
+++ b/components/Modals/NotePublishModal.tsx
@@ -5,6 +5,7 @@ import CheckBox from "~/components/Form/CheckBox";
 import FormSelect from "../Form/FormSelect";
 import Modal from "react-modal";
 import colors from "~/config/themes/colors";
+import { AuthActions } from "~/redux/auth";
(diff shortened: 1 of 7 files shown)


components/Modals/NotePublishModal.tsx
```diff
@@ -221,8 +236,8 @@ function NotePublishModal({
     let params;
     if (publishedType === "UNPUBLISHED") {
       params = {
+        assign_doi: checkBoxTwo,
```
yattias, line 239: Can we change to something like `checkBoxDOI` or something like that? I think if we switch the order of checkboxes this may cause a logic bug

yattias (commented):
Small comment. Looks good otherwise

components/Modals/NotePublishModal.tsx
thomas-vu: Good point, I'll update this.

calvinhlee23 (approved):
Code looks good!

PR #1129, opened 2022-03-22
Top Nav Redesign: (Discuss, Publish, Leaderboard)

yattias:
### What?
- Redesigned the top nav to include leaderboard, publish, discuss links while hiding others
- "Live" feed link has moved to "Recent Activity" section
- Updated feed filters design to reflect Tatiana's designs 
- Includes closed PR https://github.com/ResearchHub/researchhub-web/pull/1131

### Only three links at top
image

### Live Feed now in Latest Activity section
image

### Editors now in dropdown
image


lightninglu10:
I also think it looks a bit funky with the tab underlined and then below in the feed it also has a tab underline

lightninglu10:
I think maybe we should keep the color black in the navbar

yattias:
I added the contents of https://github.com/ResearchHub/researchhub-web/pull/1131 to this PR

PR #1131, opened 2022-03-23
Pass strings instead of Unix timestamps for time scope

yattias:
### What?
- Rather than sending unix TS, send english strings instead (i.e. today, week, ..)

### DIFF
--- a/config/utils/options.js
+++ b/config/utils/options.js
@@ -98,23 +98,28 @@ const filterOptions = [
 
 const scopeOptions = [
   {
-    value: "day",
+    value: "today",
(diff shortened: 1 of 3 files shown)


config/utils/options.js
```diff
@@ -98,23 +98,28 @@ const filterOptions = [
 
 const scopeOptions = [
   {
-    value: "day",
+    value: "today",
```
yattias, line 101: Since `scopeOptions` and its values are used by a lot of different places in the codebase. I decided to add a 
 new key: `valueForApi`.

PR #1132, opened 2022-03-25
Comment timestamps rounded incorrectly using TimeAgo

yattias:
### What?
- Fix for comment timestamps incorrectly rounding. For example, "13 days ago" would be rounded to "1 week ago".
- This leads to incorrect output and also makes it hard to debug the "discussed" filter.
- Confirmed that we do not have a function that does this in the codebase. Also confirmed that daysjs does not have a utility function as such.


thomas-vu (approved):
LGTM.

PR #1139, opened 2022-03-30
Fix issue with login on mobile

lightninglu10:
Everytime login is clicked, generate the google oauth client


calvinhlee23 (approved):
lgtm

PR #1143, opened 2022-04-01
Fixed server side rendering issue with feed

### LINKED ISSUE #1141: Feed Layout shift on page load feels "broken"
For a split second when loading the feed, the layout is not rendered correctly. This may feel broken to the user.

thomas-vu:
- Fixes #1141

### DIFF
--- a/components/UnifiedDocFeed/UnifiedDocFeedContainer.tsx
+++ b/components/UnifiedDocFeed/UnifiedDocFeedContainer.tsx
@@ -14,7 +14,7 @@ import {
   useEffectForceUpdate,
   useEffectUpdateStatesOnServerChanges,
 } from "./utils/UnifiedDocFeedUtil";
-import { ReactElement, useMemo, useRef, useState } from "react";
+import { ReactElement, useEffect, useMemo, useRef, useState } from "react";
 import { useRouter } from "next/router";
 import colors from "~/config/themes/colors";
 import CreateFeedBanner from "../Home/CreateFeedBanner";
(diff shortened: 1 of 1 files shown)


components/UnifiedDocFeed/UnifiedDocFeedContainer.tsx
```diff
@@ -248,7 +252,7 @@ function UnifiedDocFeedContainer({
           />
         </div>
       </div>
-      {Boolean(unifiedDocsLoading) ? (
+      {unifiedDocsLoading || isServer() ? (
```
yattias, line 255: Do you still need the `unifiedDocsLoading` given that you have an `isServer()` check now?

thomas-vu: Yeah.

components/UnifiedDocFeed/UnifiedDocFeedContainer.tsx
```diff
@@ -75,6 +75,10 @@ function UnifiedDocFeedContainer({
     hotV2: router.query?.hot_v2 == "true",
   };
 
+  useEffect(() => {
```
lightninglu10, line 78: can we remove this state and set it back to what we had before with the useState(false), or do we need it?

thomas-vu: We need it to avoid a strange bug with the first 3 cards. Not sure why this happens.

Screen Shot 2022-04-07 at 7 55 54 PM

yattias (approved):
LGTM

PR #1161, opened 2022-04-13
Various UI fixes

thomas-vu:


calvinhlee23 (approved):
Code looks good. Please test on staging just in case!

thomas-vu:
> Code looks good. Please test on staging just in case!

Performance seems to be roughly the same.

**With LazyLoad:**
Screen Shot 2022-04-14 at 4 18 49 PM

**Without LazyLoad:**
Screen Shot 2022-04-14 at 4 18 53 PM

Issue #1153, opened 2022-04-07
Comment field clearing on certain actions


thomas-vu:
Reported by Nick Oldberg on Slack: https://researchhub-community.slack.com/archives/C01C3M9S283/p1648484684087719

> This has been previously mentioned but I want to draw attention to it again as I see this bug adversely effecting new users to the site. 
> 
> When typing out a comment: opening up RSC withdraw/deposit window, upvoting a comment, or upvoting a post, ALL lead to whatever was written in the comment field being cleared, and there is no way to get it back. As RH cultivates deep discussion with often times lengthy comments, this is a far bigger issue than it may be on other platforms such as reddit where comments are often times brief and unsophisticated in their nature.
> 
> With the RSC withdraw/deposit window popping up on top of the normal site, there is an implied nested nature where a user could withdraw/deposit RSC and then go back to what they are doing, only to see that their comment has been cleared. I myself have accidentally upvoted a post when posting a comment, multiple times, only to have my comment completely erased and I have to type it out again. I've resorted to drafting my comment in Word and then pasting it into the comment section to avoid this, but I see this as a bit too laborious to expect new users to do

PR #1166, opened 2022-04-18
[BugFix][Temp] - Comment Editor Clearing

calvinhlee23:
#1153 - temp bugfix
- Text Editor was remounting whenever a page / thread action was taken because it would trigger discussion fetch to occur.
- I added memoization on top of TextEditor so that the component will be memoized on initial mount. 
- Note that this is temp fix. In order for this to be corrected properly, the commenting structure should be re-written both for FE & BE


yattias (approved):
@calvinhlee23 looks good. I do think we need to test on staging prior to prod deploy. LMK when it is there and I will help testing.

PR #1172, opened 2022-04-27
Calvin/censor

calvinhlee23:
Screen Shot 2022-04-27 at 1 33 30 PM


yattias:
@calvinhlee23 code looks good. 
Can we make sure that users (non-editors) can delete their own comments?

Issue #1167, opened 2022-04-20
Creating/Editing Review


yattias:
- As a user, I should be able to create a review with a `score` + `comment`
- As a user, I should be able to edit my review's comment or score

PR #1176, opened 2022-04-28
Peer Reviews

yattias:
Solves #1167 

### Creating
image

### Top
image

### Feed
image

### Discussion
image

### Profile
image

### DIFF
--- a/components/Form/ScoreInput.tsx
+++ b/components/Form/ScoreInput.tsx
@@ -8,6 +8,7 @@ type Props = {
     readOnly?: Boolean;
     overrideBarStyle?: any;
     scoreInputStyleOverride?: any;
+    withText ?: Boolean;
 };
 
 const MAX_SCORE = 10;
@@ -19,6 +20,7 @@ function ScoreInput({
   readOnly,
   overrideBarStyle = null,
   scoreInputStyleOverride = null,
+  withText = true,
 }: Props): ReactElement {
 
   const [selectedValue, setSelectedValue] = useState<number>(value);
@@ -65,7 +67,7 @@ function ScoreInput({
   return (
     <div className={css(styles.scoreInput, scoreInputStyleOverride)}>
       {barInput}
-      {Boolean(selectedValue) && 
+      {Boolean(selectedValue) && withText &&
         <div className={css(styles.scoreText)}>
           <span className={css(styles.selectedScoreText)}>{selectedValue}</span>/{MAX_SCORE}
         </div>
--- a/components/Form/ScoreInput.tsx
+++ b/components/Form/ScoreInput.tsx
@@ -1,5 +1,5 @@
 import { css, StyleSheet } from "aphrodite";
-import { ReactElement, useState } from "react";
+import { ReactElement, useEffect, useMemo, useState } from "react";
 import colors from "~/config/themes/colors";
 
 type Props = {
@@ -25,6 +25,12 @@ function ScoreInput({
 
   const [selectedValue, setSelectedValue] = useState<number>(value);
   const [hoveredValue, setHoveredValue] = useState<number>(0);
+
+  useEffect(() => {
+    if (value !== selectedValue) {
+      setSelectedValue(value);
+    }
+  }, [value])
 
   const handleSelect = (value) => {
     setSelectedValue(value);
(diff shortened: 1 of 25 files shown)


components/Form/ScoreInput.tsx
```diff
@@ -0,0 +1,150 @@
...
+      setSelectedValue(value);
+    }
+  }, [value])
+
+  const handleSelect = (value) => {
+    setSelectedValue(value);
+    onSelect && onSelect(value);
+  }
+
+  const buildBarInput = () => {
```
calvinhlee23, line 40: was there a specific reason why this was a functional? 
If not, let's just relocate this to be inside of useMemo. Or maybe making this its own component outside of  `ScoreInput` is better (I understand that keep it inside is convenient for variable access)

components/Form/ScoreInput.tsx
```diff
@@ -0,0 +1,150 @@
+import { css, StyleSheet } from "aphrodite";
```
calvinhlee23, line 1: hmm the conditionals-formatting looks off in this file. What kind of linter are you using?

components/Form/Toggle.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+function Toggle({
+  options,
+  onSelect,
+  selected,
+}: Props): ReactElement {
+
+  const handleSelect = (opt:ToggleOption) => {
+    onSelect(opt);
+  }
```
calvinhlee23, line 25: looks like this func isn't doing anything. 
Let's remove it and if we see the need to have its own handleSelect function, we can add back in(?)

components/Form/Toggle.tsx
```diff
@@ -0,0 +1,68 @@
...
+  selected: any;
+}
+
+type Props = {
+  options: Array<ToggleOption>;
+  onSelect: Function;
+  selected: any;
+};
+
+function Toggle({
```
calvinhlee23, line 17:
```suggestion
export default function Toggle({
```

components/Form/Toggle.tsx
```diff
@@ -0,0 +1,68 @@
...
+    }
+  },
+  "optionSelected": {
+    background: colors.NEW_BLUE(),
+    color: "white",
+    transition: "0.1s",
+  }
+});
+
+export default Toggle;
```
calvinhlee23, line 68:
```suggestion
```

components/Paper/Tabs/DiscussionTab.js
```diff
@@ -7,19 +7,23 @@ import Ripples from "react-ripples";
 import ReactPlaceholder from "react-placeholder";
```
calvinhlee23, line 7: Thanks for starting to refactor this. 
I've already created ticket for this for both FE & BE. 
But let's really consider migrating away from the old commenting modules.
Anytime, this component / BE needs to be updated for these, it's too much effort & I don't think it's going to be scalable the way we have them.

components/PeerReviews/PeerReviewScoreSummary.tsx
```diff
@@ -0,0 +1,92 @@
...
+  return (
+    <div className={css(styles.reviewContainer)}>
+      <span className={css(styles.reviewScoreContainer)}>
+        <span className={css(styles.reviewScore)}>{summary?.avg}</span>/10
+      </span>
+      <div className={css(styles.scoreContainer)}>
+        <span className={css(styles.dot)}>&bull;</span>
+        <ScoreInput
+          value={summary?.avg}
+          readOnly={true}
```
calvinhlee23, line 27: This evaluates to `true`
```suggestion
          readOnly
```

calvinhlee23 (changes requested):
Thoughts and comments.

components/Form/ScoreInput.tsx
yattias: eslint. not sure why it skipped. Could be that it doesn't do space formatting?
I actually don't see spaces/tabs defined in eslint

components/Form/ScoreInput.tsx
yattias: sounds reasonable. I think I had a reason earlier on.

components/Form/Toggle.tsx
yattias: yeah, I had it do something else at some point. I'll clean it up

PR #1191, opened 2022-05-25
Audit Dashboard

le-sun:
Updating query param


lightninglu10 (changes requested):
This if statement is funky -- if verdict does not exist do we still want to pass the ordering?

yattias (approved):
LGTM

PR #1199, opened 2022-06-06
Redesign of document pages

yattias:
### 1.
image

### 2.
image

### 3.
image


components/DiscussionPostMetadata.js
```diff
@@ -55,14 +55,14 @@ const DiscussionPostMetadata = (props) => {
...
     postId,
-    smaller,
     threadPath,
     toggleEdit,
     twitter,
     twitterUrl,
     username,
   } = props;
 
+  const smaller = false;
```
lightninglu10, line 65: should we remove this?

PR #1201, opened 2022-06-08
Vote grm migration

calvinhlee23:
Screen Shot 2022-06-08 at 3 02 46 PM


components/Author/Tabs/FeedCard.tsx
```diff
@@ -128,71 +136,32 @@ function FeedCard(props: FeedCardProps) {
...
+    documentID: id,
+    documentType: nullthrows(formattedDocType, "Cannot vote without doctype"),
+    onError: emptyFncWithMsg,
+    onSuccess: ({ increment, voteType }): void => {
+      setVoteState(voteType);
+      setScore(score + increment);
+    },
+    voteType: DOWNVOTE,
+  });
+  const onUpvote = createVoteHandler({
```
yattias, line 152: I recently refactored `createVoteHandler` into its own file because I wanted to remove as much as logic as possible from newly created `DocumentHeader` component. I also realized there is a lot of code around the handler that wasn't being used. 

I know this shouldn't necessarily full on you but I think it makes sense to deal with this now while you are at it. You will need to add `hypothesis` type to `getVoteUrl` function.

For reference, you should check out `DocumentHeader`

config/types/root_types.ts
```diff
@@ -10,63 +14,78 @@ export type ValueOf<ObjectType> = ObjectType[keyof ObjectType];
...
+  isRemoved?: boolean;
+};
+
+export type RhDocumentType =
+  | "eln"
+  | "hypothesis"
+  | "paper"
+  | "post"
+  | "researchhub_posts";
+export type GrmVoteType = "downvote" | "neutralvote" | "upvote";
```
yattias, line 58: I'm not sure the GRM prefix is needed. I think it will be confusing to new developers joining the team. I think the GRM part is purely a backend construct. From the FE perspective should probably just be `VoteType`

config/types/root_types.ts
```diff
@@ -10,63 +14,78 @@ export type ValueOf<ObjectType> = ObjectType[keyof ObjectType];
...
-  boostAmount: number,
-  id: ID,
+  authors: Array<AuthorProfile>;
+  score: number;
+  createdDate: string;
+  discussionCount: number;
+  unifiedDocument: UnifiedDocument;
+  hubs: Array<Hub>;
+  createdBy: CreatedBy | null;
+  userVote?: "downvote" | "upvote" | "neutralvote" | null;
```
yattias, line 24: While at it, could you update this to use the new type you created? There may be a type issue that will happen due to the `null`

config/utils/buildGrmVoteApiUri.ts
```diff
@@ -0,0 +1,46 @@
+import {
```
yattias, line 1: Can we move this to `components/Vote/api` ? It is a good pattern I believe you introduced that I have been following for a while.

config/utils/buildGrmVoteApiUri.ts
```diff
@@ -0,0 +1,46 @@
...
+    commentType?: string;
+    replyID?: string;
+    threadID?: string;
+  };
+  documentType: RhDocumentType;
+  documentID: ID;
+  voteType: GrmVoteType;
+};
+
+export function buildGrmVoteApiUri({
```
yattias, line 21: Can we update the `FLAG_GRM_CONTENT` in `api.js` to utilize this function?

components/Document/DocumentHeader.tsx
```diff
@@ -53,77 +55,55 @@ function DocumentHeader({
...
 
   let onUpvote, onDownvote;
   if (document.isReady) {
     onUpvote = createVoteHandler({
-      voteType: UPVOTE,
-      unifiedDocument,
       currentAuthor,
-      onSuccess: handleVoteSuccess,
+      currentVote: voteState?.userVote,
+      documentCreatedBy: nullthrows(createdBy),
```
yattias, line 91: We should make sure no error is thrown here if `createdBy` does not exist.
I believe this will throw an error.

We do not want this because throwing an error will likely break the header from rendering in one way or another. There are times where the user may be deleted but their paper is still available and we want to render it.

components/Vote/utils/createVoteHandler.ts
```diff
@@ -1,78 +1,99 @@
...
+};
 
 export const createVoteHandler = ({
-  voteType,
-  unifiedDocument,
+  commentPayload,
   currentAuthor,
-  selectedUserVote,
+  currentVote,
+  documentCreatedBy,
```
yattias, line 45: Can we pass in `DocumentCreatedBy` as `CreatedBy` type ? There is a parser called `parseCreatedBy` that we can use outside of this function

yattias: I think this is a good thing because we should aim to use types exclusively if possible in newly created components and functions

components/Vote/utils/createVoteHandler.ts
```diff
@@ -1,78 +1,99 @@
...
-    e.stopPropagation();
+  const formattedDocumentType =
+    documentType === "post" ? "researchhub_posts" : documentType;
+  const resolvedDocumentAuthorID =
+    documentCreatedBy?.authorProfile?.id ||
+    documentCreatedBy?.author_profile?.id;
 
-    if (!currentAuthor?.id) {
-      console.log("Not logged in");
+  return async (event: SyntheticEvent) => {
```
yattias, line 58: The changes here appear to be solid. Maybe we can test on staging first? Thanks for adding these in

components/Author/Tabs/FeedCard.tsx
yattias: NVM, I see that you are using it here. Thanks!

config/types/root_types.ts
```diff
@@ -10,63 +14,78 @@ export type ValueOf<ObjectType> = ObjectType[keyof ObjectType];
...
 export type UnifiedDocument = {
-  id: ID,
-  documentType: "post" | "paper" | "hypothesis",
-  document?: UrlDocument,
-  createdBy?: CreatedBy,
-  reviewSummary?: PeerReviewScoreSummary,
-  isRemoved: boolean,
-}
+  id: ID;
+  documentType: RhDocumentType;
```
yattias, line 69: Nice. Good change

yattias (changes requested):
I really like the refactor and think this is much needed I requested a few changes though

calvinhlee23 (commented):
Resolving some comments and replied. Still working on some comments

config/utils/buildGrmVoteApiUri.ts
calvinhlee23: I'll do it once everything starts to use VoteWidgetV2. For now, because VoteWidget is a shared component, let's keep this file as a part of utils folder. I'll make sure to move this file when the switch over is done

config/utils/buildGrmVoteApiUri.ts
calvinhlee23: I dont want to bundle this with Flag api since they are separate models with separate endpoints. I can move this to a separate file like Votes later if necessary.

components/Document/DocumentHeader.tsx
yattias: ah got it.
Yeah unfortunately we do delete users from time to time. 
There are legal laws around it.
Glad it won't break the page though.

PR #1209, opened 2022-07-06
Dropdown new post buttons

calvinhlee23:
### Code: 
- updated & added screen size handlers 
- created popover options for newPostButtons
- update killswitch for bounty questions
- update modal handling logic to reroute 

### Test: 
- Regular screen size
Screen Shot 2022-07-06 at 10 37 56 AM

- Mobile screen Size ( kept the modal to take over entire screen)
Screen Shot 2022-07-06 at 8 57 25 AM

### DIFF
--- a/components/Modals/NewPostModal.tsx
+++ b/components/Modals/NewPostModal.tsx
@@ -78,7 +78,7 @@ export const getModalOptionItems = (currentUser) => [
     key: "hypothesis",
     header: "Create a Meta-Study",
     description:
-      "Aggregate a collection of papers that support a particular scientific theory",
+      "Aggregate a collection of papers that support a particular scientific theory.",
     route: "/hypothesis/create",
     icon: (
       <HypothesisIcon
--- a/components/Modals/NewPostModal.tsx
+++ b/components/Modals/NewPostModal.tsx
@@ -18,21 +18,90 @@ import {
 } from "react";
 import { StyleSheet, css } from "aphrodite";
 import { useRouter } from "next/router";
+import {
+  DEFAULT_POST_BUTTON_VALUES,
+  NewPostButtonContext,
+  NewPostButtonContextType,
+} from "~/components/contexts/NewPostButtonContext";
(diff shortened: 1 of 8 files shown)


components/Modals/NewPostModal.tsx
```diff
@@ -18,33 +18,111 @@ import {
...
+    icon: (
+      <HypothesisIcon
+        height={40}
+        onClick={silentEmptyFnc}
+        width={40}
+        withAnimation={false}
+      />
+    ),
+  },
+  killswitch("bountyQuestion")
```
yattias, line 92: Nice use of killswitch 💯

components/Modals/NewPostModal.tsx
```diff
@@ -18,33 +18,111 @@ import {
...
+    ? {
+        key: "question",
+        header: "Ask a Question",
+        description:
+          "All posts must be scientific in nature. Ideas, theories, and questions to the community are all welcome.",
```
yattias, line 97: We will need to update the text here before go live but it's fine atm

lightninglu10: do we have new text you want to use @yattias ?

config/utils/getIsOnMobileScreenSize.ts
```diff
@@ -0,0 +1,20 @@
+import { useState } from "react";
+import { breakpoints } from "../themes/screen";
+import {
+  getCurrMediaWidth,
+  useEffectOnScreenResize,
+} from "./useEffectOnScreenResize";
+
+export function getIsOnMobileScreenSize(): boolean {
```
yattias, line 8: I would recommend we make changes that address the following:
1. This function named with `get` prefix but it also sets values. Can we rename it something else? Maybe separate setter from getter?
2. Let's avoid the name `mobile`. I think it is an industry term that we should move away because of its ambiguity. mobile could mean ipad, tablet, etc... are also mobile. Can we use `smallScreen` or even better, the breakpoint `small`

yattias (changes requested):
Small change requested

PR #1223, opened 2022-07-29
Fe mods

calvinhlee23:
### ELN
- [fix] Creating/removing a new note in ELN does not refresh sidebar
- Show title of the note in tab 
- Parse HTML in titles e.g. “&gt”

### Bounties
- Alert user if they don't have enough RSC to start bounty
- Ensure timeTo on bounty alert is correct
- Flagging for question type not working
- Add bounty constants to config file Bounty/config/constants.ts
- add utm param to twitter link

### Other
- Hide deposit feature in RSC modal
- Replace Slack Link with Discord
- Update Peer Review copy text
    1. Impact - Is the research question innovative? Do the study's findings advance the authors' field in a meaningful way?   
    2. Methods - Does the study design test the authors' hypothesis? Are the methods described in enough detail for independent replication?  
    3.  Results - Were the study's findings analyzed and interpreted reasonably ? Is the resulting data open and auditable
    4. Discussion - Do the results support the authors’ conclusions? Are there any alternative interpretations of the study's findings that the authors should have considered?


config/utils/unescapeHtmlString.ts
```diff
@@ -0,0 +1,11 @@
+import { NullableString } from "../types/root_types";
+
+export function unescapeHtmlString(htmlStr: NullableString): string {
```
yattias, line 3: Can we use a function to replace all HTML entities instead of just these? (i.e. semicolon, colon, hyphen, ..)

I think this solution would work. We should put it behind a try/catch though to avoid the feed breaking though.
https://stackoverflow.com/a/7394814/1869326

components/TextEditor/config/reviewCategories.ts
```diff
@@ -1,34 +1,38 @@
...
+      "Does the study design test the authors' hypothesis? Are the methods described in enough detail for independent replication?",
     isDefault: false,
   },
-  "results": {
+  results: {
     label: "Results",
     value: "results",
-    description: "Was the collected data analyzed and presented reasonably? Do the figures clearly represent this data?",
+    description:
+      "Were the study's findings analyzed and interpreted reasonably ? Is the resulting data open and auditable?",
```
yattias, line 26: There should be no space between "reasonably" and the first question

components/Document/DocumentActions.tsx
```diff
@@ -109,10 +110,11 @@ function DocumentActions({
             onSubmit={(flagReason, renderErrorMsg, renderSuccessMsg) => {
               flagGrmContent({
```
yattias, line 111: If possible, let's do this overriding in the API level. I have done it there thus far. This way, it's all in the same layer and if we call a function from multiple places, we can ensure it overrides it at the source

components/CKEditor/ELNEditor.js
```diff
@@ -92,10 +90,27 @@ const ELNEditor = ({
...
   const currentUserAccess = getUserNoteAccess({ user, notePerms, userOrgs });
   const noteIdLength = `${currentNote.id}`.length;
   const channelId = `${orgSlug.slice(0, 59 - noteIdLength)}-${currentNote.id}`;
+  const parsedNoteTitle = unescapeHtmlString(
+    editorInstance?.plugins?.get("Title")?.getTitle() ?? "Untitled"
+  );
+
+  useEffect(() => {
+    if (
+      !(isNullOrUndefined(typeof window) && isNullOrUndefined(typeof document))
```
yattias, line 102: This is fine but I think just checking `if (process.browser)` will do the job and a bit easier to read.
If `document` or `window` are not defined in the browser then there are big problems going on and likely other thing will break too.

yattias (changes requested):
I had a few relatively small requests

components/Document/DocumentActions.tsx
calvinhlee23: I like that. Thanks

config/utils/unescapeHtmlString.ts
calvinhlee23: This only parses numerics. I'll add on top of this. The solution is O(n). Should be fast anyway

PR #1232, opened 2022-08-09
Bounty Tracking

le-sun:
Tracking creating and awarding bounties


components/Bounty/BountyModal.tsx
```diff
@@ -95,6 +97,25 @@ function BountyModal({
     );
   };
 
+  const sendBountyCreateAmpEvent = ({ currentUser, createdBounty }) => {
```
yattias, line 100: @le-sun I think we should use the `trackEvent` function in `analytics.ts`

yattias (changes requested):

PR #1233, opened 2022-08-11
Comment bounties

lightninglu10:
Ability to add bounties to comments

Requires this PR to work: https://github.com/ResearchHub/researchhub-backend/pull/1130

UX: https://www.loom.com/share/07f0bb6753924818b07233125a18ca17


yattias (changes requested):
**QA Review**

1. Bug: Once a bounty is accepted, the headline text still reads "X is offering Y RSC" 
image

2. I think bounties should appear at the top of the discussion section and more prominently. They get lost atm.

I think just updating the colors like this would do the trick quite well

image

3. I'm pretty sure this worked before - It is not possible to accept own's answer at the moment

--Should probably do--

1. One of the core use-cases of what we are after is not entirely possible - That is, creating a bounty for a peer review. The reason it is not possible is because replies do not support "peer review". Only top-level threads do. I think we can still launch but we should really address this issue.

2. As soon as you click "Post", the badge with the bounty temporarily disappears. This may get the user a bit frightened. 


--Small--
- I think in the context of documents that are not a question the CTA should just read "Award Bounty". The badge should also read "Accepted Solution".

lightninglu10:
> --Should probably do--
> 
> 1. One of the core use-cases of what we are after is not entirely possible - That is, creating a bounty for a peer review. The reason it is not possible is because replies do not support "peer review". Only top-level threads do. I think we can still launch but we should really address this issue.
> 

Ah this is gonna be tough.. Will need to think about how to make this possible, but I think we should release without it at the moment.

yattias (changes requested):
One small comment

PR #1236, opened 2022-08-18
Types as main filters, bounty as filter option

yattias:
**What?**
- Showing types as main filtering option
- Added Bounty filter
- Update feed cards to reflect open bounty, question with answers
- New component for sort which includes time scope
- Updated feed component to support "tags" (bounty is a tag behind the scenes)

**Main changes**
image

**New component**
image


components/Document/SubmissionDetails.tsx
```diff
@@ -35,29 +39,56 @@ function SubmissionDetails({
...
   const hiddenHubs = hubs?.slice(sliceIndex) ?? [];
 
+  let authorProfile = createdBy?.authorProfile;
+  let bounty: Bounty | undefined;
+  if (bounties.length > 0) {
+    // @ts-ignore
+    authorProfile = bounties[0].createdBy?.authorProfile;
+    bounty = bounties[0];
+  }
+
```
calvinhlee23, line 49: Let's avoid variable assignments like these. It's really hard to read after some time passes. Also, pretty sure no need for ts-ignore
```suggestion
   const bounty =  (bounties ?? [])[0]?.createdBy?.authorProfile;
  }
```

components/Document/SubmissionDetails.tsx
```diff
@@ -133,6 +174,22 @@ const styles = StyleSheet.create({
...
+  dotWithMargin: {
+    marginLeft: 5,
+  },
+  rscIcon: {
+    verticalAlign: "text-top",
+    marginLeft: 5,
+  },
+  rscText: {
+    fontWeight: 600,
+    color: colors.ORANGE_DARK2(),
```
calvinhlee23, line 186: I'm not a fan of these empty function calls for hard string literals. It's fine for now. but let's just avoid these all together.

components/UnifiedDocFeed/FeedOrderingDropdown.tsx
```diff
@@ -0,0 +1,166 @@
...
+  const selectedOrderingObj = sortOpts.find(
+    (o) => o.value === selectedOrderingValue
+  );
+  const dropdownRef = useRef<HTMLDivElement>(null);
+
+  const _handleOrderingClick = (selected) => {
+    onOrderingSelect(selected);
+    if (selected.disableScope) {
+      setIsOpen(false);
+    }
```
calvinhlee23, line 24:
```suggestion
      setIsOpen(selected.disableScope);
 ```

File: components/UnifiedDocFeed/FeedOrderingDropdown.tsx [hunk 4]
```diff
@@ -0,0 +1,166 @@
...
+  );
+}
+
+const styles = StyleSheet.create({
+  feedOrderingDropdown: {
+    position: "relative",
+  },
+  display: {
+    display: "flex",
+    // color: colors.BLACK(0.6),
```
calvinhlee23, line 103:
```suggestion
```

components/UnifiedDocFeed/FeedOrderingDropdown.tsx
```diff
@@ -0,0 +1,166 @@
...
+  },
+  displayDown: {},
+  dropdownBody: {
+    position: "absolute",
+    top: 25,
+    right: 0,
+    zIndex: 50,
+    height: "auto",
+    width: 200,
+    background: "#fefefe",
```
calvinhlee23, line 121: colors?

config/utils/getUnifiedDocType.ts
```diff
@@ -69,6 +71,8 @@ export function getBEUnifiedDocType(
       return "posts";
     case "hypothesis":
       return "hypothesis";
+    case "bounties":
+        return "all";
     case "paper":
       return "paper";
     default:
```
calvinhlee23, line 78:
```suggestion
      return "hypothesis";
    case "paper":
      return "paper";
    case "bounties":
    default:
```

calvinhlee23 (changes requested):
Thanks for updating the feedMenu component. I made some comments that are easy to fix

calvinhlee23:
Please resolve merge conflicts before merging

components/Document/SubmissionDetails.tsx
yattias: Thanks for the suggestion. It doesn't quite accomplish what I was trying to do (notice there are two vars assigned) but I get the sentiment of it and will update

components/UnifiedDocFeed/FeedOrderingDropdown.tsx
yattias: ah yes

components/UnifiedDocFeed/FeedOrderingDropdown.tsx
yattias: I decided to keep things as is because the right line would actually be

```
 setIsOpen(!selected.disableScope);
 ```
 
 instead of 
```
 setIsOpen(selected.disableScope);
 ```

Even though it is an extra two lines, having the if statement there makes it a lot easier for me to read.
My brain takes a while figuring out negation and how it relates to an op like `setIsOpen`

calvinhlee23:
LGTM. Please test feature changes on staging & resolve conflicts against master before merging.

calvinhlee23:
LGTM. Please test feature changes on staging & resolve conflicts against master before merging.

calvinhlee23 (approved):

PR #1240, opened 2022-08-29
New Filtering Mechanism

yattias:
- New filtering mechanism based on Figma designs
- Filtering state derived from URL params
- Removed sidebar "My hubs" / "All hubs" in favor of adding it to main filtering section
- Added "Tags" subfilters which is commented out atm until BE changes are ready
- Refactor menu components into multiple sub-components for ease of use
- Refactor mechanism of fetching docs from BE

image


components/UnifiedDocFeed/constants/UnifiedDocFilters.ts
```diff
@@ -15,55 +29,139 @@ export const UnifiedDocFilterLabels = {
...
+    availableFor: [
+      feedTypeOpts["all"].value,
+      feedTypeOpts["paper"].value,
+      feedTypeOpts["post"].value,
+      feedTypeOpts["question"].value,
+      feedTypeOpts["meta-study"].value,
+      feedTypeOpts["bounty"].value,
+    ],
+  },
+  // {
```
yattias, line 75: Temporarily commented out until BE is ready

components/Hubs/MyHubsDropdown.tsx
```diff
@@ -0,0 +1,120 @@
...
+        </a>
+      </Link>
+    );
+  };
+
+  const hubOpts = useMemo(() => {
+    const myHubsOpts = (hubState.subscribedHubs || []).map((h) =>
+      renderDropdownOpt(h)
+    );
+    return myHubsOpts;
```
calvinhlee23, line 41: Explicit nullcheck. "Or" operator is based on boolean. 0 / null / empty strings / undefined are all falsey. Better to be type-specific. Prevents bugs this way.
```suggestion
    return (hubState.subscribedHubs ?? []).map((h) =>
      renderDropdownOpt(h)
    );
```

components/ResearchHubBanner.js
```diff
@@ -70,12 +71,11 @@ class ResearchHubBanner extends Component {
...
                 ? "ResearchHub"
                 : this.props.hub.name}
-              !
             </span>
           </div>
           <div className={css(styles.subtext, styles.text)}>
-            We're a community seeking to improve prioritization, collaboration,
-            reproducibility, and funding of scientific research.{" "}
+            We are bulding an open platform and community whose goal it is to
+            accelerate science.{" "}
```
calvinhlee23, line 78: Having strings represented within react components like below preserves formatting. Just a thing to consider
```suggestion
            {"We are bulding an open platform and community whose goal it is to
            accelerate science. "} 
```

components/UnifiedDocFeed/FeedMenu/FeedMenu.tsx
```diff
@@ -0,0 +1,236 @@
...
+      query: router.query,
+      pathname: router.pathname,
+    });
+  }, [router.pathname, router.query]);
+
+  useEffect(() => {
+    useEffectForOutsideMenuClick({
+      setTagsMenuOpenFor,
+    });
+  }, []);
```
calvinhlee23, line 37: implementation of this call looks a little fishy. Could you double check

components/UnifiedDocFeed/FeedMenu/FeedMenu.tsx
```diff
@@ -0,0 +1,236 @@
...
+    const tabsAsHTML = tabs.map((tabObj) => {
+      if (tabObj.value === selectedFilters.type) {
+        // @ts-ignore
+        tabObj.isSelected = true;
+      }
+      return tabObj;
+    });
+
+    return tabsAsHTML;
+  };
```
calvinhlee23, line 73: let's move this call outside of the component. No need to redefine this function & calling it everytime the component mounts. Also makes it easier to read.

components/UnifiedDocFeed/FeedMenu/FeedMenu.tsx
```diff
@@ -0,0 +1,236 @@
...
+      <FeedMenuTab
+        selectedFilters={selectedFilters}
+        tabObj={t}
+        handleOpenTagsMenu={(forType) => setTagsMenuOpenFor(forType)}
+        handleFilterSelect={(selected) =>
+          handleFilterSelect({ router, ...selected })
+        }
+        isTagsMenuOpen={tagsMenuOpenFor === t.value}
+      />
+    ));
```
calvinhlee23, line 87: Unnecessary variable assignment
```suggestion
    return getTabFilters({ selectedFilters }).map((t) => (
      <FeedMenuTab
        selectedFilters={selectedFilters}
        tabObj={t}
        handleOpenTagsMenu={(forType) => setTagsMenuOpenFor(forType)}
        handleFilterSelect={(selected) =>
          handleFilterSelect({ router, ...selected })
        }
        isTagsMenuOpen={tagsMenuOpenFor === t.value}
      />
    ));
```

components/UnifiedDocFeed/FeedMenu/FeedMenu.tsx
```diff
@@ -0,0 +1,236 @@
...
+      selectedOrderingValue={selectedFilters.sort}
+      selectedScopeValue={selectedFilters.time}
+      onOrderingSelect={(selected) =>
+        handleFilterSelect({ router, sort: selected.value })
+      }
+      onScopeSelect={(selected) =>
+        handleFilterSelect({ router, timeScope: selected.value })
+      }
+    />
+  );
```
calvinhlee23, line 102: Looks like there's no need to assign this element to a named variable. 
Could we paste it within `return`?

components/UnifiedDocFeed/FeedMenu/FeedMenuMobileScrollControls.tsx
```diff
@@ -0,0 +1,97 @@
...
+import useEffectForHorizontalScroll from "../utils/useEffectForHorizontalScroll";
+import { css, StyleSheet } from "aphrodite";
+import icons from "~/config/themes/icons";
+
+const FeedMenuMobileScrollControls = ({ tabsContainerRef, viewportWidth }) => {
+  const [showMobileLeftScroll, setShowMobileLeftScroll] = useState(false);
+  const [showMobileRightScroll, setShowMobileRightScroll] = useState(false);
+  const [showMobileScrollNav, setShowMobileScrollNav] = useState(false);
+
+  useEffect(() => {
```
calvinhlee23, line 11: same here. usage of useEffect within useEffect looks very weird.

components/UnifiedDocFeed/FeedMenu/FeedMenuTab.tsx
```diff
@@ -0,0 +1,130 @@
...
+  tabObj,
+  handleOpenTagsMenu,
+  handleFilterSelect,
+  isTagsMenuOpen,
+}: Args) => {
+  const router = useRouter();
+  const isSelected = tabObj.value === selectedFilters.type;
+  const nestedOptions = tagFilters.filter((sub) =>
+    sub.availableFor.includes(tabObj.value)
+  );
```
calvinhlee23, line 32: Can we revisit this approach here? 
There's really no need for a rendering / child component to have been passed down a full tabObj to determine whether or not it's been selected. I think boolean variable might be sufficient in this case.

components/UnifiedDocFeed/FeedMenu/FeedMenuTab.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+  const url = buildTypeFilterUrl({ tabObj, router });
+  return (
+    <div
+      className={`${css(
+        styles.tab,
+        tabObj.isSelected && styles.tabSelected
```
calvinhlee23, line 39: This looks off here. If tabObj already has `isSelected` attr, why do we need to check if the tab is selected above?

components/UnifiedDocFeed/FeedMenu/FeedMenuTopLevelFilters.tsx
```diff
@@ -0,0 +1,213 @@
...
+  feedOrderingElem: React.ReactNode;
+};
+
+const FeedMenuTopLevelFilters = ({
+  selectedFilters,
+  currentUser,
+  hubState,
+  feedOrderingElem,
+}: Args) => {
+  const router = useRouter();
```
calvinhlee23, line 27: Please use `getCurrentUser()` instead of using redux export directly.

components/UnifiedDocFeed/UnifiedDocFeedContainer.tsx
```diff
@@ -23,58 +20,45 @@ import FeedBlurWithButton from "./FeedBlurWithButton";
...
-    scope: scopeOptions[0],
-    tags: [] as any,
-  });
-
+  const selectedFilters = useMemo(() => {
+    return getSelectedUrlFilters({
+      query: router.query,
+      pathname: router.pathname,
+    });
+  }, [router.pathname, router.query]);
```
calvinhlee23, line 47: Pretty sure there's no need to memoize filters here. 
Updating path will trigger everything within feed to rerender because we are changing the fetch to solely depend on the URL now.

components/UnifiedDocFeed/utils/getSelectedUrlFilters.ts
```diff
@@ -0,0 +1,80 @@
...
+  tagFilters,
+  topLevelFilters,
+} from "../constants/UnifiedDocFilters";
+import { scopeOptions } from "~/config/utils/options";
+
+export type SelectedUrlFilters = {
+  topLevel: string | undefined;
+  type: string | undefined;
+  sort: string | undefined;
+  time: string | undefined;
```
calvinhlee23, line 13: Please use `NullableString` unless we wanted to make use of `undefined` specifically

components/UnifiedDocFeed/utils/getSelectedUrlFilters.ts
```diff
@@ -0,0 +1,80 @@
...
+    type: Object.values(feedTypeOpts)[0].value,
+    sort: sortOpts[0].value,
+    time: scopeOptions[0].value,
+    tags: <string[]>[],
+  };
+  const selected = { ...defaults };
+
+  if (!(query || pathname)) {
+    return defaults;
+  }
```
calvinhlee23, line 32: This will prevent false positives on falsy JS values
```suggestion
  if (!(isEmpty(query ?? pathname)) {
    return defaults;
  }
```

components/UnifiedDocFeed/utils/getSelectedUrlFilters.ts
```diff
@@ -0,0 +1,80 @@
...
+  query,
+  pathname,
+}): SelectedUrlFilters => {
+  const defaults = {
+    topLevel: topLevelFilters[0].value,
+    type: Object.values(feedTypeOpts)[0].value,
+    sort: sortOpts[0].value,
+    time: scopeOptions[0].value,
+    tags: <string[]>[],
+  };
```
calvinhlee23, line 27: keep it outside the component.

components/UnifiedDocFeed/utils/getSelectedUrlFilters.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  if (!(query || pathname)) {
+    return defaults;
+  }
+
+  if (Array.isArray(query.tags)) {
+    selected.tags = [...query.tags];
+  } else if (query.tags) {
+    selected.tags.push(query.tags);
+  }
```
calvinhlee23, line 38: obj mutation like this makes it hard to understand what's happening. Could we inline this at line 28 such that
```
const selected = { 
...defaults, 
tags: Array.isArray(query.tags) ? [...query.tags] : selected.tags.push(query.tags)
}
```

components/UnifiedDocFeed/utils/getSelectedUrlFilters.ts
```diff
@@ -0,0 +1,80 @@
...
+  const foundSort = sortOpts.find((opt) => opt.value === query?.sort)?.value;
+  const foundTopLevelFilter = topLevelFilters.find(
+    (f) => f.url === pathname
+  )?.value;
+  const foundTypeFilter = Object.values(feedTypeOpts).find(
+    (opt) => opt.value === query?.type
+  )?.value;
+  const foundTimeScope = scopeOptions.find(
+    (opt) => opt.value === query?.time
+  )?.value;
```
calvinhlee23, line 49: This segment is kinda very ugly. 
Do we really need to sort & find what's there within this component rather than having something passed down?

calvinhlee23 (changes requested):
Went through about half. More review to come. Meanwhile,  requestion changes

components/Hubs/MyHubsDropdown.tsx
yattias: Thanks for the suggestion.
I try to avoid using `??` unless I am working with values that require it. For example, if I were working with an integer that evaluated to 0.  I prefer continued use of the `||` operator as it makes me think of types more carefully and makes the code a bit easier to read since the double pipe exists in most languages.

LMK if you think my statement can fail due to a value I am not taking into account.

components/UnifiedDocFeed/FeedMenu/FeedMenu.tsx
yattias: Good call

components/UnifiedDocFeed/FeedMenu/FeedMenu.tsx
yattias: Good call

components/UnifiedDocFeed/FeedMenu/FeedMenu.tsx
yattias: I was using it in a couple of places but just consolidated. Yes.

components/UnifiedDocFeed/FeedMenu/FeedMenuMobileScrollControls.tsx
yattias: The implementation is solid but you questioning of it made me feel that it is not clear so I renamed things and aliased.

This is responsible for determining whether or not to display the mobile scroll (see right arrow)

image

components/UnifiedDocFeed/FeedMenu/FeedMenuTab.tsx
yattias: Yeah, that's fair. I just updated.

components/UnifiedDocFeed/UnifiedDocFeedContainer.tsx
yattias: You know what, you're right. 

I initially had a different implementation and this is an artifact of it

components/UnifiedDocFeed/utils/getSelectedUrlFilters.ts
yattias: `defaults` is an object returned as part of the function. Moving it outside of this component will result in polluting the global scope.

It is true that node probably avoids such things however it feels dirty placing a var returning within a function in the global scope (i.e. outside of the function definition).

components/UnifiedDocFeed/utils/getSelectedUrlFilters.ts
yattias: Cool. Didn't really know about this type

components/UnifiedDocFeed/utils/getSelectedUrlFilters.ts
yattias: I actually want the `||` opeartor here. I want `""` to be falsy.

Though, LMK if you see some way which the above will fail

components/UnifiedDocFeed/utils/getSelectedUrlFilters.ts
yattias: The goal of this function is to generate an object from URL state. If we pass down things to it then it prevents the logic from being centralized.

There is something I could potentially though which is to update the structure I store config in `UnifiedFeedFilters`. I'll see if that's feasible

components/Hubs/MyHubsDropdown.tsx
calvinhlee23: For this speicific example, it works. 
But, there's good use case for being specific with null / falsey values, especially when it comes to upserting data.
Not the ideal situation, but as a programmer, where null & false values make huge differences. 
Let's take RSC as an example: 
Account value of `undefined` or `null` maybe

components/UnifiedDocFeed/utils/getSelectedUrlFilters.ts
calvinhlee23: in case query = 0 & pathname  = ""
- !(query || pathname) => !(false || false) => evaluates to true 
- isEmpty(query ?? pathname) =>  !isEmpty(0) => evalues to false

Above can lead to weird bugs. Former requires URL to be always be presented as intended. Latter disregards any edgecase

components/Hubs/MyHubsDropdown.tsx
yattias: You make a good point. I would actually want the application to break if the value were non-array.
So 0.map would cause the app the break and us to question what's going on. I like it.

calvinhlee23:
lgtm

calvinhlee23 (approved):

PR #1243, opened 2022-08-31
[HomeRevamp][RightSidebar][1/n][mashed 4 Prs] - Code Migration 

calvinhlee23:
- https://github.com/ResearchHub/researchhub-web/pull/1242/files#diff-464ebf8d0c7e6224e7fd371fb94cc535c74dbf460d678237a4fcb3d3248223ae
- RightSidebar Setup
- calls to open bounties
- Carousel setup. 
- Sectional setup

Screen Shot 2022-08-31 at 9 01 51 AM


components/Bounty/api/fetchOpenBounties.ts
```diff
@@ -0,0 +1,55 @@
...
+export type SimpleBounty = {
+  amount: string;
+  content_type: any /* intentional any */;
+  created_by: User;
+  expiration_date: string;
+  id: ID;
+  item: any /* intentional any */;
+};
+
+export const fetchOpenBounties = ({ onError, onSuccess, page }: Args): void => {
```
yattias, line 22: A couple of things. 

1. We already have a function called `fetchBountiesAPI` which does just that. I think you can tweak it to return a particular status
2. We should use the Bounty class as a type instead of something like the type defined above. If it doesn't do what you need, you can tweak it. You should be able to do `bounties.map(b => new Bounty(b))`
3. The Bounty class currently holds API logic + data. I think we should fetch through a method defined on the class to simplify things on the FE. I think it would be nice to do something like `Bounty.fetchOpen()` and have the method return a list of Bounty objects. That object already has three API endpoints it hits

components/Home/sidebar/HomeSidebarBountiesSection.tsx
```diff
@@ -0,0 +1,138 @@
...
+    paginationInfo,
+    setOpenBounties,
+    setPaginationInfo,
+  });
+
+  const { isFetching, page = 1 } = paginationInfo;
+  const isReadyToRender = !isFetching && page > 0;
+  const _isLoadingMore = !isFetching && page !== 1;
+
+  const bountyItems = openBounties?.map(
```
yattias, line 61: Most of the logic in this function is handled by the `Bounty` class.

documents, amounts, expiration date formatting, ...

components/Home/sidebar/HomeSidebarBountiesSection.tsx
```diff
@@ -0,0 +1,138 @@
...
+      <ReactPlaceholder
+        ready={isReadyToRender}
+        customPlaceholder={
+          <HubEntryPlaceholder
+            color={colors.PLACEHOLDER_CARD_BACKGROUND}
+            rows={3}
+          />
+        }
+      >
+        {bountyItems.slice(0, 4)}
```
yattias, line 134: Small: Can we put the 4 into a constant/var ?

components/Home/sidebar/HomeSidebarFeaturedDocsSection.tsx
```diff
@@ -0,0 +1,130 @@
...
+        relatedDocID ??
+        unified_document?.documents?.id ??
+        (unified_document?.documents ?? [])[0]?.id;
+      const resolvedSlug =
+        slug ??
+        unified_document?.documents?.slug ??
+        (unified_document?.documents ?? [])[0]?.slug;
+
+      return (
+        <BountiesSidebarItem
```
yattias, line 92: For all bounty related things, we should just pass the bounty object instead of all the misc attributes. See function `BountyAlert` for example

components/Home/sidebar/sidebar_items/BountiesSidebarItem.tsx
```diff
@@ -0,0 +1,118 @@
...
+  bountiesSidebarItemContainer: {
+    borderLeft: "2px solid transparent",
+    color: colors.TEXT_GREY(1),
+    cursor: "pointer",
+    padding: "12px 0",
+    textDecoration: "none",
+    width: "100%",
+    ":hover": {
+      background: colors.LIGHT_GREY_BACKGROUND,
+      borderLeft: "2px solid blue",
```
yattias, line 79: I think the default BLUE is probably not consistent with the RH blue

components/Home/sidebar/sidebar_items/BountiesSidebarItem.tsx
```diff
@@ -0,0 +1,118 @@
...
+}: Props): ReactElement {
+  const roundedOfferAmount = bountyAmount;
+
+  return (
+    <div className={css(styles.bountiesSidebarItemContainer)}>
+      {/* NOTE: href is subject to change */}
+      <a
+        style={{ textDecoration: "none" }}
+        href={`/${
+          documentType === "question" ? "post" : documentType
```
yattias, line 36: There is a method `getUrlToUniDoc` in routing.js which does this. I think you should tweak it a bit to inclue your use case

components/Home/sidebar/sidebar_items/BountiesSidebarItem.tsx
calvinhlee23: Fixed on `rightSidebarMigrate4`

components/Home/sidebar/sidebar_items/BountiesSidebarItem.tsx
calvinhlee23: Commented on this on `rightSidebarMigrate3`. 
The object payload from the BE doesn't necessarily have `UniDoc`.  May need to update BE or come up with another function.

components/Home/sidebar/HomeSidebarBountiesSection.tsx
calvinhlee23: Done in `rightSidebarMigrate4`

components/Home/sidebar/HomeSidebarFeaturedDocsSection.tsx
calvinhlee23: I'm seeing TS errors related to `Bounty` objects & implemented code including `amount` being passed as number or string depending on the context. I would like to also avoid using heavy duty conversion when the BE payload is very simple (which in this case is)

PR #1248, opened 2022-09-01
Group Bounties

lightninglu10:
Allowing users to add to bounties

### DIFF
--- a/components/Bounty/BountyAlert.tsx
+++ b/components/Bounty/BountyAlert.tsx
@@ -95,9 +95,6 @@ const BountyAlert = ({
                 onClickOutside={(_event) => setPopoverOpen(false)}
                 targetContent={
                   <span
-                    onMouseEnter={() => {
-                      setPopoverOpen(!popoverOpen);
-                    }}
                     onClick={() => {
                       setPopoverOpen(!popoverOpen);
                     }}
--- a/components/Bounty/BountyAlert.tsx
+++ b/components/Bounty/BountyAlert.tsx
@@ -24,7 +24,6 @@ const BountyAlert = ({
     return a.id - b.id;
   });
 
-  console.log(allBounties);
   let amount = 0;
   if (bounty) {
     timeRemaining = bounty.timeRemaining;
--- a/components/Bounty/BountyAlert.tsx
+++ b/components/Bounty/BountyAlert.tsx
@@ -4,6 +4,8 @@ import ALink from "../ALink";
 import Bounty, { BOUNTY_STATUS } from "~/config/types/bounty";
 import colors from "~/config/themes/colors";
 import numeral from "numeral";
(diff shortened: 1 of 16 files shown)


components/Paper/Tabs/DiscussionTab.js
```diff
@@ -196,7 +196,9 @@ const DiscussionTab = (props) => {
                       key={`thread-${t.data.id}`}
                       {...t.data}
                       bounties={
-                        bountyMap[t.data.id]
+                        props.bounties && props.bounties.length
```
lightninglu10, line 199: This needs to be updated. Kind of messy right now, but we have two bounty types:

1. Questions
2. Comments

We're using one field right now "bounties" for both of them. props.bounties is from Questions and the bountyMap and t.data.bounties is for comment bounties

calvinhlee23 (approved):
lgtm

components/PostPageCard.js
```diff
@@ -123,9 +123,9 @@ class PostPageCard extends Component {
             document={post}
             onDocumentRemove={removePost}
             onDocumentRestore={restorePost}
-            hasBounties={this.props.hasBounties || this.props.bounty}
-            bounty={this.props.bounty}
-            allBounties={this.props.allBounties}
+            hasBounties={this.props.hasBounties || this.props.bounties}
+            allBounties={this.props.bounties}
+            bountyType="question"
```
yattias, line 128: See my first comment above

components/Bounty/BountyAlert.tsx
```diff
@@ -8,9 +8,14 @@ import numeral from "numeral";
 type BountyAlertParams = {
   bounty: Bounty;
   allBounties: [Bounty];
+  bountyType: string;
```
yattias, line 11: I spent a bunch of time curating the `Bounty` class into an object that contains functionality and data related for all things bounty.

I think it would be really super to add an attribute called "type" which can be "question" or "bounty". This type doesn't have to be a BE thing, just a FE thing to make things easier to use.

components/Bounty/BountyAlert.tsx
```diff
@@ -23,7 +28,8 @@ const BountyAlert = ({ bounty, allBounties }: BountyAlertParams) => {
       return null;
     }
   } else if (allBounties.length) {
-    const firstBounty = new Bounty(allBounties[0]);
+    const firstBounty =
+      bountyType === "question" ? allBounties[0] : new Bounty(allBounties[0]);
```
yattias, line 32: There is probably a use case but this feels weird because the same array can sometimes have bounties and other times raw data? Is it possible to covert `allBounties` to always be an array of bounty objects earlier on?

yattias (changes requested):
A few small change requests.

PR #1251, opened 2022-09-07
Update style for cards with bounties

yattias:
- Updated style for cards w/bounties
- No longer rendering title/body of comment if comment bounties exist

image

### DIFF
--- a/config/types/bounty.ts
+++ b/config/types/bounty.ts
@@ -51,7 +51,7 @@ export default class Bounty {
     this._timeRemainingInDays = timeToInUnits({ date: raw.expiration_date, unit: "day" });
     this._createdBy = parseCreatedBy(raw.created_by);
     this._amount = parseFloat(raw.amount);
-    this._formattedAmount = parseInt(this._amount.toFixed(0)).toLocaleString();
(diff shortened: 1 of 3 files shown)


lightninglu10 (approved):

config/types/bounty.ts
```diff
@@ -50,6 +51,7 @@ export default class Bounty {
     this._timeRemainingInDays = timeToInUnits({ date: raw.expiration_date, unit: "day" });
     this._createdBy = parseCreatedBy(raw.created_by);
     this._amount = parseFloat(raw.amount);
+    this._formattedAmount = parseInt(this._amount.toFixed(0)).toLocaleString();
```
lightninglu10, line 54: I don't like parsing int like this -- what if the actual number is something that isn't an int? I think we should change

lightninglu10: One big thing to consider is that RSC has 18 decimals. So whenever we work with RSC we should always allow for the possibility of non integers (but our inputs can force integers, just when we read things back they should always allow for non integers)

yattias: Yeah makes sense.

PR #1263, opened 2022-09-19
[Home][Revamp] - LeftSidebar Full Migration

calvinhlee23:
- Increase pagination size
- LeftSizebar root position
- Other navigation tools ui update 
- Logic clean up
- Transition / screensize handling for navigators 
- animation 

TEST: 

https://user-images.githubusercontent.com/17622930/191135106-ca6f044b-c8a3-43f9-a4ff-753cfe9c123a.mov


components/Home/sidebar/RootLeftSidebar.tsx
```diff
@@ -0,0 +1,436 @@
...
+      icon: icons.squares,
+      label: "Hubs",
+      isActive: ["/hubs"].includes(pathname),
+      isMinimized,
+      onClick: (event: SyntheticEvent): void => {
+        event.preventDefault();
+        router.push("/hubs");
+      },
+    },
+    isLoggedIn
```
yattias, line 64: We should be showing the notebook even if the user is not logged in. See what we do currently in the top nav "Menu" link and duplicate that logic. Reason being is that we want to signal to users that we have that feature

components/Navbar.js
```diff
@@ -399,6 +342,13 @@ const Navbar = (props) => {
     setSideMenu(state.isOpen);
   }
 
+  const pathname = router?.pathname ?? "";
+  const headerLabel = pathname.includes("notebook")
+    ? "Lab Notebook"
+    : pathname.includes("leaderboard")
+    ? "Leaderboard"
+    : deSlug(router?.query?.slug ?? "");
```
yattias, line 350: We shouldn't need this. We only need to show title for the things on the left sidebar.
I like the approach though

yattias (approved):

PR #1317, opened 2023-03-01
Foundation for new comment system

yattias:
- Foundation of new comment system
- Includes a proof of concept for parsing backend responses into a typescripted ES6 class


components/Comment/lib/CommentModel.ts
```diff
@@ -0,0 +1,100 @@
+import { formatDateStandard, timeSince } from "~/config/utils/dates";
```
calvinhlee23, line 1: I'm not a big fan of having classes just to format data from backend to serve FE purposes. 
I doesn't really do much more than a simple object, and classes are harder to maintain

calvinhlee23 (changes requested):

PR #1318, opened 2023-03-04
Comment system v2 foundation

yattias:
UI is very much in progress.  The  focus on this PR is on the code/data structures and data flow.

**Kinda looks like this at the moment:**
image

UI aside, this PR includes foundation for our comment system. Take a look at  the directory/file structure as well


components/Comment/lib/types.ts
```diff
@@ -0,0 +1,53 @@
+import { parseCreatedBy } from "~/config/types/contribution";
+import { CreatedBy, ID } from "~/config/types/root_types";
+import { formatDateStandard, timeSince } from "~/config/utils/dates";
+
+export enum POST_TYPES {
```
calebds, line 5: Seems like an opportunity to rely on a common definition with https://github.com/ResearchHub/researchhub-web/blob/35cdceb27280516a454ccd8736b5fb52a38cf4f9/components/TextEditor/config/postTypes.ts#L15.

components/Comment/lib/types.ts
```diff
@@ -0,0 +1,53 @@
...
+  ANSWER = "ANSWER",
+}
+
+export type Comment = {
+  id: ID;
+  createdDate: string;
+  updatedDate: string;
+  timeAgo: string;
+  createdBy: CreatedBy | null;
+  content: any;
```
calebds, line 18: `string | undefined` might be a safer bet here, looking at how this field is used

components/Comment/lib/isQuillEmpty.ts
```diff
@@ -0,0 +1,7 @@
+export default function isQuillEmpty(content) {
```
calebds, line 1: functional nit: `return JSON.stringify(content) === '{"ops":[{"insert":"\\n"}]}'`

components/Comment/lib/findComment.ts
```diff
@@ -0,0 +1,54 @@
...
+      return found;
+    }
+  }
+};
+
+const findComment = ({
+  id,
+  comments,
+}: findCommentArgs): FoundComment | undefined => {
+  for (let i = 0; i < comments.length; i++) {
```
calebds, line 45: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find can be a nice functional alternative

components/Comment/Comment.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+type CommentArgs = {
+  comment: CommentType;
+  handleUpdate: Function;
+  handleCreate: Function;
+};
+
+const Comment = ({ comment, handleUpdate, handleCreate }: CommentArgs) => {
+  const [isReplyOpen, setIsReplyOpen] = useState(false);
+  const [isEditMode, setIsEditMode] = useState(false);
```
lightninglu10, line 18: Can we type these to be Boolean? 

We've run into issues before where if this changes to a number, specifically 0, then the below 

```
isReplyOpen && ...
```

Will display a 0 on the website

components/Comment/CommentEditor.tsx
```diff
@@ -0,0 +1,133 @@
...
+      setIsSubmitDisabled(true);
+    } else {
+      setIsSubmitDisabled(false);
+    }
+
+    _setContent(value);
+  };
+
+  useEffect(() => {
+    const _handleClick = (e) => {
```
lightninglu10, line 46: This whole section may be needed -- without diving into the logic a bit more it's hard for me to tell but on initial glance it looks kind of opaque and may be hard to maintain.

Just marking as a section that could be bug prone

components/Comment/CommentFeed.tsx
```diff
@@ -0,0 +1,83 @@
...
+};
+
+const CommentFeed = ({ unifiedDocumentId }: Args) => {
+  const [comments, setComments] = useState<CommentType[]>([]);
+  const [isFetching, setIsFetching] = useState(true);
+
+  useEffect(() => {
+    const _fetchComments = () => {
+      setIsFetching(true);
+      fetchCommentsAPI({ unifiedDocumentId }).then(
```
lightninglu10, line 25: I wonder -- should we move away from .then syntax and instead 

```
const _fechComments = async () => {
     ...
     const comments = await fetchCommentsAPI(...);
     setComments(comments);
     setIsFetching(false);
}
```

Something like this, just a thought

components/Comment/Comment.tsx
yattias: Sure, I didn't think it was needed to type case primitives but this makes sense

components/Comment/CommentEditor.tsx
yattias: Yeah I agree, I kinda don't like it. 

I  consolidated this in the other PR. LMK if you have thoughts about how to make it even cleaner https://github.com/ResearchHub/researchhub-web/blob/81e4a51a13dc97e960418cabb55b210121c6211d/components/Comment/CommentEditor.tsx#L48

components/Comment/lib/findComment.ts
yattias: Thanks for the comment. The reason for the custom find here is the recursive nature of the search operation. Each comment object has `children` which in turn can have their own child comments and so on...

`findComment` is the exported "public" method and the `_find` is supposed to be the helper recursive call

components/Comment/lib/isQuillEmpty.ts
yattias: 👍

components/Comment/lib/types.ts
yattias: I appreciate the thoroughness in the PR and noticing there are two similar type defs. You are absolutely right!

What I have been doing here is copying all the good stuff from the "old comment" system into then new `Comment/*` comment system because the intention is to delete the old one once the new one is ready. This means that we will temporarily have a few duplicate things here and there

components/Comment/lib/types.ts
yattias: You know what, this should have been `object` type. It was my bad defaulting it to a string. I updated it in my follow-up PR branch https://github.com/ResearchHub/researchhub-web/blob/ad700c21f819a126c39cb02a902c73c6e6b10a58/components/Comment/CommentEditor.tsx#L15-L22

components/Comment/CommentFeed.tsx
yattias: I actually prefer async/await syntax since it avoids [pyramid of doom](https://en.wikipedia.org/wiki/Pyramid_of_doom_(programming)). I updated in subsequent PR https://github.com/ResearchHub/researchhub-web/blob/aeec3fa455750c4b63ab57a166ba63f16aa2904d/components/Comment/CommentFeed.tsx#L37

PR #1319, opened 2023-03-06
Comment foundation, part II

yattias:
- `CommentTypeSelector` component (peer review, discussion, ..)
- `useQuill` instead of `react-quill`
- Changed `CreatedBy` type to `RHUser` and `parseCreatedBy` to `parseUser` (`RHUser` and not `User` because we already have a `User` type that seems to be used in various places - legacy and should not be used going forward)
- Async/Await instead of `then` in various api places


components/Comment/lib/quill.css
```diff
@@ -0,0 +1,308 @@
+.ql-container {
```
yattias, line 1: This file is copy of old quill file. Going to make changes and do not want  to affect old comment module

components/Comment/CommentEditor.tsx
```diff
@@ -68,34 +68,50 @@ const CommentEditor = ({
...
+        }
+        else {
+          setIsSubmitDisabled(false);
+        }
+      });
     }
-  }, [_isFocused]);
+  }, [quill]);
+
+  if (Quill && !quill) {
```
yattias, line 87: This if statement will change to something better

components/Comment/hooks/useQuill.ts
```diff
@@ -0,0 +1,109 @@
+// Imported from https://github.com/gtgalone/react-quilljs
```
yattias, line 1: This file is imported from https://github.com/gtgalone/react-quilljs

components/Comment/CommentEditor.tsx
```diff
@@ -68,34 +68,50 @@ const CommentEditor = ({
...
+    const MagicUrl = require('quill-magic-url').default;
```
lightninglu10, line 88: This section feels a bit weird to me. 

I think we currently do something like this -- possibly to get around server side rendering, but is there another way to do it?

Dynamic imports?

components/Comment/CommentEditor.tsx
```diff
@@ -1,62 +1,62 @@
...
     const _handleClick = (e) => {
-      const isOutsideClick =
-        !_isPreviewRef.current && !editorRef.current?.contains(e.target);
-      const excludedElems = [".reply-btn", ".edit-btn"];
-      const clickOnExcluded = excludedElems.reduce(
-        (prev, selector) => Boolean(prev || e.target.closest(selector)),
-        false
-      );
+      const isFullToolbarTriggerClick = e.target.closest(".show-full-editor");
+      const isFullToolbarClick = e.target.closest(".ql-full-editor");
```
lightninglu10, line 52: Is it possible to get this inside the react components instead of building handler here? 

For example in a <div> below, if we can do 

```
<div onClick onEnter onLeave

```

Or something like that instead. Not sure it's possible, just wondering

components/Comment/CommentEditor.tsx
yattias: Yeah I need to experiment a bit

components/Comment/lib/types.ts
```diff
@@ -35,8 +34,8 @@ export const parseComment = ({ raw, parent }: parseCommentArgs): Comment => {
     createdDate: formatDateStandard(raw.created_date),
     updatedDate: formatDateStandard(raw.created_date),
     timeAgo: timeSince(raw.created_date),
-    createdBy: parseCreatedBy(raw.created_by),
-    content: raw.content,
+    createdBy: parseUser(raw.created_by),
```
yattias, line 37: @lightninglu10 Just FYI, I  changed the name of this mainly because I think I picked a wrong name.
There are instances where I need to pass in the currently logged in user and its a bit awkward to do `parseCreatedBy(currentUser)`. More intuitive to do `parseUser(currentUser)`

yattias: For simplicity, let's aim to merge this PR first today.

config/utils/isOutsideClick.ts
```diff
@@ -0,0 +1,17 @@
...
+  const isWithin = el === clickedEl || el?.contains(clickedEl);
+  const clickOnExcluded = exclude.reduce(
+    (prev, selector) => Boolean(prev || clickedEl.closest(selector as keyof HTMLElementTagNameMap)),
+    false
+  );
+
+  return !(isWithin || clickOnExcluded);
+}
+
+export default isOutsideClick;
```
yattias, line 17: @calvinhlee23 / @lightninglu10 just FYI - Created this helper function since we do this operation quite often

components/Comment/lib/commentTypes.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { COMMENT_TYPES } from "./types";
+
+const commentTypes = [
+  {
+    label: "Discuss",
+    verb: "discussing",
+    value: COMMENT_TYPES.DISCUSSION,
+    isDefault: true,
+    group: "contribute",
+    icon: <FontAwesomeIcon icon={faComments} />,
```
yattias, line 12: @calvinhlee23 / @lightninglu10 As we discussed in Colorado, this is the right way to import font awesome icons. It is most performant according to their docs.

components/Comment/CommentEditorToolbar.tsx
```diff
@@ -0,0 +1,76 @@
...
+    document.addEventListener("click", _handleClick);
+
+    return () => {
+      document.removeEventListener("click", _handleClick);
+    };
+  }, []);
+
+  return (
+    <div id={editorId} className="ql-toolbar" ref={toolbarRef}>
+      <span className="ql-formats">
```
yattias, line 37: `className` need to be as such in order for quill to work properly

components/Comment/hooks/useQuill.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const [isLoaded, setIsLoaded] = useState(false);
+  const [obj, setObj] = useState({
+    Quill: undefined as any | undefined,
+    quillRef,
+    quill: undefined as Quill | undefined,
+    editorRef: quillRef,
+    editor: undefined as Quill | undefined,
+  });
+
+  useEffect(() => {
```
yattias, line 89: Verified that Quill gets mounted once and never remounted

components/Comment/CommentEditor.tsx
yattias: Ended up moving this logic to `useQuill`.
We actually want to avoid dynamic imports here since it causes a race condition. The creator of this hook recommends avoiding them.

`CommentEditor` should be cleaner. 
Verified Quill does not get remounted intermittently - only once. It now fees more lightweight.

components/Comment/CommentEditor.tsx
yattias: I ended up doing something like this by moving the state of the toolbar to within the toolbar. Code is cleaner now on both `CommentEditor` and `CommentEditorToolbar`

components/Comment/CommentEditorToolbar.tsx
calvinhlee23: can you leave this comment within code.

config/utils/isOutsideClick.ts
calvinhlee23: Can you have a callback as an arg, `onClickOutside` and make this into a ueffect so we don't have to redefine it at the component level?

calvinhlee23 (changes requested):
Looks good sofar. 
I made a comment on `isOutsideClick` helper function.

lightninglu10 (approved):

PR #1322, opened 2023-03-13
[ReferenceManager][1/n] - Migrate existing page from other repo to a renderable stage.

calvinhlee23:
### Code: 
- code source main: https://github.com/ResearchHub/researchhub-citation-manager 
- most of the code from the other repo is written by me & was reviewed before merging. Quality should be good.
- [subsequent branches [2-5 & rebase]/n are merged here

### Test: 
Screen Shot 2023-03-13 at 4 07 04 PM


components/ReferenceManager/form/PrimaryButton.tsx
```diff
@@ -0,0 +1,37 @@
+import { ReactElement } from "react";
+import { Container, Box, Button } from "@mui/material";
```
yattias, line 2: The recommended way to import is using tree shaking for reduced bundle size:

```
import Container from "@mui/material/Container";
import Box from "@mui/material/Box";
```

....

components/ReferenceManager/references/ReferencesContainer.tsx
```diff
@@ -0,0 +1,117 @@
+import {
+  Box,
+  IconButton,
+  InputAdornment,
+  Typography,
+  OutlinedInput,
+} from "@mui/material";
```
yattias, line 7: Same comment as above

components/ReferenceManager/shared/basic_page_layout/BasicTogglableNavbarLeft.tsx
```diff
@@ -0,0 +1,341 @@
...
+    duration: theme.transitions.duration.leavingScreen,
+  }),
+  overflowX: "hidden",
+  width: `calc(${theme.spacing(7)} + 1px)`,
+  [theme.breakpoints.up("sm")]: {
+    width: `calc(${theme.spacing(8)} + 1px)`,
+  },
+});
+
+const doNotPassProps = new Set(["isOpen", "navWidth"]);
```
yattias, line 93: I'm not sure what the purpose of this is

components/ReferenceManager/shared/basic_page_layout/BasicTogglableNavbarLeft.tsx
```diff
@@ -0,0 +1,341 @@
...
+          cursor: "pointer",
+          display: "flex",
+          justifyContent: "space-between",
+          width: "100%",
+          paddingLeft: isOpen ? "22px" : "16px",
+          paddingRight: "16px",
+        }}
+      >
+        {isOpen && (
+          <Image
```
yattias, line 160: Thanks for using `next/Image`

components/ReferenceManager/shared/menu/DropdownMenu.tsx
yattias, line 1: I realize that the structure of this app started as a separate app so I'm sure things were a bit different in that conext.

In this context, I think we should remove the `shared` folder since it is a bit misleading and have:

`ReferenceManager/Menu/DropdownMenu.tsx` (notice the case)

yattias (commented):
I really appreciate the usage of typescript.
I think we should walk through the PRs and discuss my comments just so we are all on the same page and have a deeper understanding of the code.

components/ReferenceManager/form/PrimaryButton.tsx
lightninglu10: I feel like webpack does automatic tree shaking so it shouldn't matter, but I might be wrong here.

PR #1325, opened 2023-03-17
[ReferenceManager][3/n] - Fetch & Dynamic Mapping 

calvinhlee23:
Screen Shot 2023-03-17 at 9 53 32 AM


components/ReferenceManager/references/api/fetchCurrentUserReferenceCitations.ts
yattias, line 1: I'm trying to standardize our app structure. Can we put all api related things in `ReferenceManager/lib/api.ts` ?

lightninglu10: We should talk about it more, but I actually think that approach makes it hard to pick out files. 

Like in the editor if you're trying to find api.ts, 10 of them will show up and you'd have to specify ReferenceManager/lib/api.ts directly which feels a bit more cumbersome. 

I think naming things differently allows for looking for a specific file easier.

PR #1327, opened 2023-03-21
Comment System v2, Part IV

yattias:
- Sidebar component
- Toggle sidebar component
- Read more/less comment
- Vote component
- Sort

image


components/Comment/lib/quill.css
yattias, line 1: Css imported from old quill css file and slightly modified

components/Bounty/CreateBountyBtn.tsx
yattias, line 1: The only modification is wrapping of `<IconButton>` around the clickable are

components/Comment/CommentActions.tsx
```diff
@@ -1,25 +1,92 @@
...
+          <CommentVote
+            comment={comment}
+            score={comment.score}
+            userVote={comment.userVote}
+            documentType={document.documentType}
+            documentID={document.id}
+          />
+        </div>
+        <div className={`${css(styles.action)} reply-btn`}>
+          <IconButton onClick={() => null}>
```
lightninglu10, line 44: Do we need the onClick here?

components/Comment/CommentActions.tsx
```diff
@@ -1,25 +1,92 @@
...
+            </span>
+          </IconButton>
         </div>
-        <div className={`${css(styles.action)} edit-btn`}>
-          {/* TODO: This requires updating font awesome common types */}
-          <FontAwesomeIcon icon={faEdit} />
-          <span onClick={() => handleEdit()}>Edit</span>
+        {currentUser?.id === comment.createdBy.id && (
+          <div className={`${css(styles.action, styles.editAction)} edit-btn`}>
+            <IconButton onClick={() => null}>
```
lightninglu10, line 61: Same as above, any reason to have onClick here?

components/Comment/CommentActions.tsx
```diff
@@ -1,25 +1,92 @@
...
+              <FontAwesomeIcon
+                icon={faPencil}
+                style={{ color: colors.secondary.text, fontSize: 18 }}
+              />
+              <span
+                className={css(styles.actionText)}
+                onClick={() => handleEdit()}
```
lightninglu10, line 68: Same as above

components/Comment/CommentActions.tsx
```diff
@@ -1,25 +1,92 @@
...
+              >
+                Edit
+              </span>
+            </IconButton>
+          </div>
+        )}
+        <div className={`${css(styles.action, styles.actionReply)} reply-btn`}>
+          <IconButton onClick={() => null}>
```
lightninglu10, line 76: ^

components/Comment/CommentActions.tsx
```diff
@@ -1,25 +1,92 @@
...
+            <Image
+              src="/static/icons/reply.png"
+              height={14}
+              width={17}
+              alt="Reply"
+            />
+            <span
+              className={css(styles.actionText)}
+              onClick={() => handleReply()}
```
lightninglu10, line 85: ^

components/Comment/CommentActions.tsx
```diff
@@ -36,6 +103,15 @@ const styles = StyleSheet.create({
...
     cursor: "pointer",
+    fontSize: 15,
+    fontWeight: 500,
+  },
+  actionReply: {
+    marginLeft: "auto",
+  },
+  editAction: {},
+  actionText: {
+    color: colors.secondary.text,
```
lightninglu10, line 114: We should update this later, but I think we need to get rid of our big config objects. Because it's an object, all the colors are imported at this import which is causing the bundle size to grow, when we only use this one color here. 

Also, I think it's better if we make these functions so we can pass opacity into the color, or make a separate getter function like 

```
color({opacity: .5, color: secondaryTextColor});
```

and have the default opacity be 1.

components/Comment/CommentAvatars.tsx
```diff
@@ -0,0 +1,34 @@
+import AuthorFacePile from "../shared/AuthorFacePile";
+import { css, StyleSheet } from "aphrodite";
+import { AuthorProfile, ID } from "~/config/types/root_types";
+
+type CommentAvatarsArgs = {
```
lightninglu10, line 5: Should we move this out of this file, or leave it for simplicity? 

I'm a fan of leaving things that don't need to be exported in one file so you don't need to open 2 files to do the same operation.

components/Comment/CommentEditor.tsx
```diff
@@ -53,17 +59,40 @@ const CommentEditor = ({
     content,
   });
 
+  if (previewWhenInactive) {
+    useEffectHandleClick({
+      el: editorRef.current,
+      onOutsideClick: () => {
+        setIsPreviewMode(true);
+        isPreviewModeRef.current = true;
```
lightninglu10, line 67: Is this ref needed? I'm not seeing it used anywhere

components/Comment/CommentEditor.tsx
```diff
@@ -53,17 +59,40 @@ const CommentEditor = ({
...
+      },
+    });
+  }
+
   useEffect(() => {
-    const isDisabled = isQuillEmpty(_content) ? true : false;
-    setIsSubmitDisabled(isDisabled);
+    const isEmpty = isQuillEmpty(_content) ? true : false;
+    setIsEmpty(isEmpty);
+    isEmptyRef.current = isEmpty;
```
lightninglu10, line 80: Is this ref also being used? Not seeing it being invoked anywhere

components/Comment/CommentEditor.tsx
```diff
@@ -74,7 +103,12 @@ const CommentEditor = ({
         )}
         <div className={css(styles.editor)}>
           <div ref={quillRef} />
-          <div className={css(styles.toolbarContainer)}>
+          <div
+            className={css(
+              styles.toolbarContainer,
+              isPreviewMode && styles.hidden
```
lightninglu10, line 109: Instead of hiding, with display hidden, should we instead conditionally render this section?

```
isPreviewMode && ...
```

components/Comment/CommentFeed.tsx
```diff
@@ -17,33 +17,55 @@ import CommentFilters from "./CommentFilters";
...
-  unifiedDocumentId: ID;
+  document: TopLevelDocument;
+  WrapperEl: any;
 };
 
-const CommentFeed = ({ unifiedDocumentId }: Args) => {
+const CommentFeed = ({ document, WrapperEl = React.Fragment }: Args) => {
   const [comments, setComments] = useState<CommentType[]>([]);
-  const [isFetching, setIsFetching] = useState<boolean>(true);
+  const [isFetching, setIsFetching] = useState<boolean>(false);
```
lightninglu10, line 31: We have 3 fetch render states here. Is it possible to boil it down to just 1? Feels harder to maintain with 3 states that might be able to be set with 1.

components/Comment/CommentSidebarToggle.tsx
```diff
@@ -3,39 +3,76 @@ import colors from "./lib/colors";
...
-  const bountyDisplayVal = (bountyAmount > 1000 ? bountyAmount / 1000 : bountyAmount).toFixed(0);
+  const bountyDisplayVal =
+    bountyAmount > 1000
+      ? (bountyAmount / 1000).toFixed(0) + "k"
+      : bountyAmount.toFixed(0);
 
-  if (!isReady) {
-    return null;
-  }
+  const [mountEl, setMountEl] = useState<HTMLElement | null>(null);
```
lightninglu10, line 29: Instead of setting this in the state, should we do it in a ref instead? 

So the component won't re-render on set of the mountEl

components/Comment/CommentSidebarToggle.tsx
```diff
@@ -3,39 +3,76 @@ import colors from "./lib/colors";
...
+  useEffect(() => {
+    if (!mountEl) {
+      const _mountEl = document.getElementById(config.toggle.elemToMountAt);
```
lightninglu10, line 32: This feels like a weird way to write react. Is there a way we can do this without trying to get the element by Id and then conditionally calling createPortal?

components/shared/AuthorFacePile.tsx
```diff
@@ -62,6 +62,9 @@ export default function AuthorFacePile({
         Boolean(horizontal) && styles.horizontal,
         overrideStyle
       )}
+      // When negative margins are used, the width of this container will be smaller than actually is and result in unintended spacing of adjacent elements.
```
lightninglu10, line 65: I've seen us have these problems with a lot of other components. Feels like we need to clean up some css in some spots so these weird spacings don't happen

components/Comment/CommentActions.tsx
yattias: Yeah, this is is WIP. My next PR will have something in there. Just needed a placeholder now.

components/Comment/CommentActions.tsx
```diff
@@ -1,25 +1,92 @@
...
+          <IconButton onClick={() => null}>
+            <Image
+              src="/static/icons/tip.png"
+              height={20}
+              width={21}
+              alt="Tip"
+            />
+            <span
+              className={css(styles.actionText)}
+              onClick={() => handleReply()}
```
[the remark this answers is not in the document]
yattias: This is is WIP. My next PR will have something in there. Just needed a placeholder now.

components/Comment/CommentActions.tsx
yattias: This is is WIP. My next PR will have something in there. Just needed a placeholder now.

components/Comment/CommentActions.tsx
yattias: This is is WIP. My next PR will have something in there. Just needed a placeholder now.

components/Comment/CommentActions.tsx
yattias: This is is WIP. My next PR will have something in there. Just needed a placeholder now.

components/Comment/CommentAvatars.tsx
yattias: You know, you are probably right. Initially this file was `CommentAuthors` which had more stuff in it

components/Comment/CommentActions.tsx
yattias: I do agree that our colors need to be reworked from the ground-up at some point. Ideally we should have a single config entity. 

As for whether it should be an object or function. Not sure. We will need to run some tests. You may be onto something here but not sure if an object adds non-trivial bloat. It may though.

I know in the case of icons it does because many of these icons are quite big. I did a quick test where icons.js was updated to export an empty object literal `{}` instead of one with icons and that reduce the "shared" bundle size by 120kb from 880kb => 760kb.

components/Comment/CommentEditor.tsx
```diff
@@ -53,17 +59,40 @@ const CommentEditor = ({
     content,
   });
 
+  if (previewWhenInactive) {
```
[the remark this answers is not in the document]
yattias: Ah I spent a bunch of time thinking this one through. I think it would be best to walk through this in a huddle.

components/Comment/CommentEditor.tsx
yattias: It is used. Let me huddle you and walk through this.

components/Comment/CommentEditor.tsx
yattias: It is used. Let me huddle you and walk through this.

components/Comment/CommentEditor.tsx
yattias: That was my initial thought but conditionally rendering this results in a choppy quill toolbar rendering when `isPreviewMode == false`

components/Comment/CommentActions.tsx
lightninglu10: Yeah, we should do it like FontAwesome where it's just named exports, like 

`export secondaryTextColor = ...`

Our API is also a massive one where we'll probably save over 100kb in bundle size just by converting it to named exports.

Maybe a job for GPT4? lol

PR #1332, opened 2023-03-29
Comment System v2, Part V

yattias:
**Updated usage of new comment feed to be as such**

```
<CommentFeed document={document} context="sidebar" previewModeAsDefault={false} />

context: "sidebar" | "drawer" | null
```

**Due to the recursive nature of the comment system, refactored into the following component structure**

`<CommentFeed />`
 &nbsp;    |_ `<CommentList />` <- new component
 &nbsp; &nbsp; &nbsp; |_ `<Comment />`
 &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;|_ `<CommentList comments={comment.children} />`
 &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;|_ `<Comment />`
 &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;|_ ...

**Added `CommentTreeContext`**
- Is used in various levels of the tree to refer to state of the tree
- Only `<CommentFeed />` will hold the state of the tree. Other components will refer to its source of truth through context

**Load more capability at different levels of the tree**
- Not just on top level, but also on subtrees

**Drawer component for comments in mobile devices**
`<CommentDrawer />`

### DIFF
--- a/components/Comment/Comment.tsx
+++ b/components/Comment/Comment.tsx
@@ -126,11 +126,8 @@ const Comment = ({
           />
         </div>
       )}
-      {comment.children.length > 0 && (          
-        <div className={css(styles.children)}>
-          <CommentList parentComment={comment} comments={comment.children} document={document} />
-        </div>
-      )}
+
+      <CommentList parentComment={comment} comments={comment.children} document={document} />
     </div>
   );
 };
@@ -141,12 +138,6 @@ const styles = StyleSheet.create({
   },
   headerWrapper: {
     marginBottom: 10,
-  },
-  children: {
-    marginLeft: 10,
-    paddingLeft: 15,
-    borderLeft: `3px solid ${colors.border}`,
-    paddingTop: 15,
   },
   editorWrapper: {
     marginTop: 15,
--- a/components/Comment/Comment.tsx
+++ b/components/Comment/Comment.tsx
@@ -9,7 +9,6 @@ import { TopLevelDocument } from "~/config/types/root_types";
 import colors from "./lib/colors";
 import { hasOpenBounties } from "./lib/bounty";
 import Button from "../Form/Button";
-import CreateBountyBtn from "../Bounty/CreateBountyBtn";
 
 type CommentArgs = {
   comment: CommentType;
@@ -40,6 +39,15 @@ const Comment = ({
       </div>
     ));
   }, [comment.children]);
+
+  const _handleToggleReply = () => {
+    if (isReplyOpen && confirm("Discard changes?")) {
+      setIsReplyOpen(false);
+    }
+    else {
+      setIsReplyOpen(true);
+    }
+  }
 
   return (
     <div>
@@ -88,17 +96,21 @@ const Comment = ({
         </div>
         <div className={css(styles.actionsWrapper)}>
           <CommentActions
-            handleReply={() => setIsReplyOpen(!isReplyOpen)}
+            toggleReply={() => _handleToggleReply()}
             document={document}
             comment={comment}
+            isReplyOpen={isReplyOpen}
           />
         </div>
       </div>
       {isReplyOpen && (
-        <CommentEditor
-          handleSubmit={({ content }) => handleUpdate({ content })}
-          editorId={`reply-to-${comment.id}`}
-        />
+        <div className={css(styles.editorWrapper)}>
+          <CommentEditor
+            handleSubmit={({ content }) => handleCreate({ content })}
+            editorId={`reply-to-${comment.id}`}
+            placeholder={`Enter reply to this comment`}
+          />
+        </div>
       )}
       {comment.children.length > 0 && (
         <div className={css(styles.children)}>{_childCommentElems}</div>
@@ -119,6 +131,9 @@ const styles = StyleSheet.create({
     paddingLeft: 15,
     borderLeft: `3px solid ${colors.border}`,
     paddingTop: 15,
+  },
+  editorWrapper: {
+    marginTop: 15,
   },
   actionsWrapper: {},
   mainWrapper: {},
--- a/components/Comment/Comment.tsx
+++ b/components/Comment/Comment.tsx
@@ -127,7 +127,12 @@ const Comment = ({
         </div>
       )}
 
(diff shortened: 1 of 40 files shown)


components/Comment/CommentEditor.tsx
```diff
@@ -80,9 +94,45 @@ const CommentEditor = ({
     isEmptyRef.current = isEmpty;
   }, [_content]);
 
+  const _handleSubmit = async () => {
+    setIsSubmitting(true);
+    try {
+      // if (quill!.getLength() <= config.comment.minLength) {
```
yattias, line 100: Please ignore

components/Comment/Comment.tsx
```diff
@@ -54,13 +61,21 @@ const Comment = ({
...
+              handleSubmit={async (args) => {
+                console.log("args", args);
+                await handleUpdate(args);
+                setIsEditMode(false);
+              }}
               content={comment.content}
+              commentId={comment.id}
+              author={currentUser?.authorProfile}
               editorId={`edit-${comment.id}`}
+              handleClose={() => _handleCloseEdit()}
```
lightninglu10, line 78: in these cases, if we aren't passing any params, can we pass it like this instead:

```
handleClose={_handleCloseEdit}
```

components/Comment/Comment.tsx
```diff
@@ -87,22 +102,38 @@ const Comment = ({
         </div>
         <div className={css(styles.actionsWrapper)}>
           <CommentActions
-            handleEdit={() => setIsEditMode(!isEditMode)}
-            handleReply={() => setIsReplyOpen(!isReplyOpen)}
+            toggleReply={() => _handleToggleReply()}
```
lightninglu10, line 105: same as above

components/Comment/Comment.tsx
```diff
@@ -87,22 +102,38 @@ const Comment = ({
...
-          handleSubmit={({ content }) => handleUpdate({ content })}
-          editorId={`reply-to-${comment.id}`}
-        />
-      )}
-      {comment.children.length > 0 && (
-        <div className={css(styles.children)}>{_childCommentElems}</div>
+        <div className={css(styles.editorWrapper)}>
+          <CommentEditor
+            handleClose={() => _handleToggleReply()}
+            handleSubmit={async ({ content, commentType }) => {
```
lightninglu10, line 116: Can we define this function above and not inline?

components/Comment/CommentDrawer.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+const CommentDrawer = ({ children, comments, isInitialFetchDone }) => {
+  const [isOpen, setIsOpen] = useState<boolean>(false);
+  const [drawerEl, setDrawerEl] = useState<null | HTMLElement>(null);
+  const openBountyAmount = comments.reduce(
+    (total, comment) => total + getBountyAmount({ comment }),
+    0
+  );
+
+  const commentCount = 999;
```
lightninglu10, line 28: Do we want this set here like this?

components/Comment/CommentEditor.tsx
```diff
@@ -80,9 +92,44 @@ const CommentEditor = ({
...
+  };
+
   return (
-    (<div ref={editorRef} className={css(styles.commentEditor)}>
+    <div ref={editorRef} className={css(styles.commentEditor)}>
       <div>
+        {handleClose && (
+          <IconButton
+            overrideStyle={styles.closeBtn}
+            onClick={() => handleClose()}
```
lightninglu10, line 128: pass function without calling

components/Comment/CommentEditor.tsx
```diff
@@ -114,36 +163,47 @@ const CommentEditor = ({
         </div>
       </div>
       <div className={css(styles.actions)}>
+        <div style={{ width: 70 }}>
+          <Button
+            fullWidth
+            label={
+              isSubmitting ? (
+                <Loader color="white" type="clip" size={22} />
```
lightninglu10, line 171: maybe use ClipLoader here instead

components/Comment/CommentEmptyState.tsx
```diff
@@ -0,0 +1,46 @@
...
+type Args = {
+  forSection: NullableString;
+  documentType: RhDocumentType;
+  height: string;
+};
+
+const CommentEmptyState = ({ forSection, documentType, height }: Args) => {
+  return (
+    <div className={css(styles.wrapper)} style={{ height }}>
+      {forSection === "REVIEW" ? (
```
lightninglu10, line 14: Anyway we can split this a bit better? I always find this style of switching between `forSection` a bit hard to read / maintain.

Maybe it's easier to read with a switch statement? Or instead have the parent component pass in the render state

components/Comment/CommentList.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+  const fetchMore = async ({}) => {
+    setIsFetchingMore(true);
+    try {
+      const response = await fetchCommentsAPI({
+        documentId: document.id,
+        documentType: document.documentType,
+        sort: commentTreeState.sort,
+        filter: commentTreeState.filter,
+        page: currentPage,
```
lightninglu10, line 58: This feels like we'll be fetching a dupe page on the first load of fetchMore, but I might not be seeing it properly

components/Comment/hooks/useQuillContent.ts
```diff
@@ -1,29 +1,45 @@
...
-}
+  quill: Quill | undefined;
+  content: object;
+};
 
 const useQuillContent = ({ quill, content = {} }: Args) => {
-
   const [_content, _setContent] = useState<object>(content);
-  const contentRef = useRef<object>(_content);
+  const [isInitialized, setIsInitialized] = useState<boolean>(false);
```
lightninglu10, line 12: instead of using a state object here, maybe it's better to use a ref

```
const isInitialized = useRef(false)`

...
if (!isInitialized.current) {
...
}
isInitialized.current = true;
```

PR #1346, opened 2023-04-12
Dismissable Feature React hook

yattias:
- New hook to allow for dismissing things like alert, tooltips and other things which checks backend as well
- Changed `useEffectNewFeature` to `useDismissableHook` and updated functionality a bit
- New functionality includes status `"unchecked" |  "checking" | "checked"` to allow components to decide whether they want to display the dismissable thing by default or not
- Optimized hook so it is only called once
- Updated `ExitableBanner` to utilize hook


components/Tooltips/NewFeatureTooltip.tsx
```diff
@@ -17,30 +15,24 @@ function NewFeatureTooltip({
...
-    featureName,
-  });
-  const normalizedFeatureName = featureName.toLocaleLowerCase();
-
-  const handleDismiss = () => {
-    window.localStorage.setItem(
-      `feature_${normalizedFeatureName}_clicked`,
-      "true"
-    );
+  // Kobe: Turning this feature off temporarily because something broke with it.
```
yattias, line 18: This whole file needs to be re-examined and updated to use the new hook. Until we have another use case, this will continue to be commented out

config/hooks/useDismissableFeature.ts
yattias, line 1: Please read my PR info. This is an update to the existing hook. I'm changed things up and added slightly more functionality

components/Banner/ExitableBanner.tsx
```diff
@@ -27,13 +29,16 @@ export default function ExitableBanner({
...
+  
+  const {
+    isDismissed,
+    dismissFeature,
+    dismissStatus
+  } = useDismissableFeature({ auth, featureName: bannerKey })
+  
 
-  if (isExited) {
+  if (dismissStatus === "unchecked" || (dismissStatus === "checked" && isDismissed)) {
```
calvinhlee23, line 41: lol my first reaction to this was "wuut" haha can you explain the second part of the conditional?

components/Tooltips/NewFeatureTooltip.tsx
calvinhlee23: yeah I agree. I think the updates we've made recently with packages messed with how some of the useEffect methods were being used.

config/hooks/useDismissableFeature.ts
```diff
@@ -0,0 +1,90 @@
...
+type UseEffectNewFeatureArgs = {
+  auth?: any;
+  featureName: string;
+};
+
+type PostNewFeatureNotifiedArgs = {
+  auth?: any;
+  featureName: string;
+};
+
```
calvinhlee23, line 14:
```suggestion
type DissMissStatus = "unchecked" | "checked" | "checking"
```

config/hooks/useDismissableFeature.ts
```diff
@@ -0,0 +1,90 @@
...
+export function useDismissableFeature({
+  auth,
+  featureName,
+}: UseEffectNewFeatureArgs): {
+  isDismissed: boolean,
+  dismissFeature: Function,
+  dismissStatus: "unchecked" | "checked" | "checking"
```
calvinhlee23, line 21:
```suggestion
  dismissStatus: DissMissStatus
```

config/hooks/useDismissableFeature.ts
```diff
@@ -0,0 +1,90 @@
...
+} {
+  const [isDismissed, setIsDismissed] = useState<boolean>(false);
+  const [status, setStatus] = useState<"unchecked" | "checking" | "checked" >("unchecked");
```
calvinhlee23, line 24:
```suggestion
  const [status, setStatus] = useState<DissMissStatus >("unchecked");
```

config/hooks/useDismissableFeature.ts
calvinhlee23: This way gives better type hints & less prone to bugs

calvinhlee23 (approved):
Please take a look at the comments i've made before merging

components/Banner/ExitableBanner.tsx
yattias: Ah right, `isDismissed` is `false` by default until `dismissStatus`  is fetched from server.

Basically the component does not render the `ExitableBanner` until confirmed by the server that user has not dismissed it

PR #1404, opened 2023-05-04
[Document V3] Part I: Structure and Foundation

yattias:
- Directory Structure for document page and tabs within
- Posts, Questions, Papers share the same document page "Wrapper"
  - Dynamic route: `[documentType]/[documentId]/[documentSlug]/[tabName]`
- Emphasis is on fast, static generation of document page


pages/_app.js
```diff
@@ -182,9 +182,9 @@ MyApp.getInitialProps = async (appContext) => {
     "/hubs/[slug]",
     "/hubs/[slug]/live",
     "/paper/[paperId]/[paperName]",
+    "/doc-v2/[documentType]/[documentId]/[documentSlug]",
```
yattias, line 185: `doc-v2` is a temporary route which will be removed once we are ready to make the switch

PR #1410, opened 2023-05-04
Standalone Login Page

### LINKED ISSUE #1401: Nextjs Error: "Repeated forward-slashes (//) or backslashes \ are not valid in the href."

### LINKED ISSUE #1408: Standalone Login Page
The primary use case for this is when user is logged out and they attempt to access a page that requires them to be authenticated. E.g. Notebook, Reference Manager, ..

In these cases, we should redirect them to a login page. Once they login, we can gracefully redirect them back to the page they attempted to access.

### LINKED ISSUE #1417: Use Cookies instead of LocalStorage for auth
We currently use both localStorage and cookies to determine the auth status of a user. This causes inconsistency between FE and BE

yattias:
Fixes #1401 
Fixes #1408 
Fixes #1417 

### Standalone Login page
- To use, pass `redirect` param to redirect to specific page after login
e.g.
```
  if (!authToken) {
    return {
      redirect: {
        destination: `/login?redirect=${ctx.req.url}`,
        permanent: false,
      },
    };
  }
```  
https://www.loom.com/share/1b0eb9a6d20745f4acdf5e4e6bcf28ee

### Double Slash issue

**Description of fix**
Rather than hard-coding the exact notebook URL in the left nav (which varies from user to user), we just redirect user to "/notebook" which in turn redirects to appropriate notebook

**Benefits:**
- No more "//" error
- New `/notebook` route is easy to remember and jump back into
- Central logic for determining which notebook to redirect to. This way, in various component we just need to redirect to "/notebook" rather than duplicating the logic in multiple places. 
- Central error handling in case user does not have default notebook

### Auth Changes

Wasn't planning on it initially, but after further investigation, ended up updating all auth code to utilize cookies instead of localStorage


pages/notebook/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+  // Ideally, we should redirect user to login page if they don't have an auth token
+  // However, because we do not have such a page, we will redirect them to 404
+  // Open issue: https://github.com/ResearchHub/researchhub-web/issues/1408
+  if (!authToken) {
+    return {
+      props: {
+        errorCode: 404,
+      },
+    };
+  }
```
yattias, line 37: @lightninglu10 wanted to bring your attention to the above scenario ⬆️ 
LMK if you have thoughts about a standalone login page.

yattias: I think the right thing to do here will be to wrap the login modal in a page. That should not be difficult.
If it's cool, I'd like to merge this and tackle the login modal separately.

yattias: Actually, I'll just do it as part of this PR. I think it is okay.

yattias: Done

pages/notebook/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  return null;
+};
+
+export async function getServerSideProps(ctx) {
+  const cookies = nookies.get(ctx);
+  const authToken = cookies[AUTH_TOKEN];
+
+  if (!authToken) {
```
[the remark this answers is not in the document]
yattias: How/when do you think it will happen? I'm not sure I understand the scenario that might result in it failing.
Could there be a scenario where user is logged in but cookies are not set?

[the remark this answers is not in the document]
yattias: I see. Do you think it is best then to make a request on the BE to some endpoint like the one that gets user info and then examine if it results in a 401?

yattias:
Ended up doing more work here than I initially set out to do.
Important to know:

1. Updated auth code flows to use cookies instead of localStorage. That way, we can safely determine the auth status of a user on both backend and frontend
2. In order to achieve the above, I needed to import `@quantfive/js-web-config` package so that I could update the underlying mechanism to not use localStorage
3. Created a `ensureAuthenticated` helper function that can be used in any page that uses `getServerSideProps` in order to ensure user is logged in prior to viewing page.

As a side effect, the issue where user intermittently logs out should be now fixed

config/api/api.js
```diff
@@ -0,0 +1,173 @@
+import Cookies from "js-cookie";
```
yattias, line 1: Imported from "import { API } from "@quantfive/js-web-config"

config/api/api.js
yattias, line 1: Imported from "import { API } from "@quantfive/js-web-config"

config/api/helpers.js
yattias, line 1: Imported from "import { API } from "@quantfive/js-web-config"

config/auth/ensureAuthenticated.ts
```diff
@@ -0,0 +1,80 @@
+import nookies from "nookies";
```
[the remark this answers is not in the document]
yattias: This thought occurred to me as well.
It appears that nookies is a library that allows reading of cookies in a nextjs specific way whereas the other package is agnostic of the underlying framework (but can't be easily used in server-side).

The right solution is to get rid of `js-cookie` and replace with `nookies`. I agree with consolidation but would rather punt on it for now for reason:
1. It appears the bundle size of `js-cookie` is 780B (less than 1kb)
2. Tested everything thoroughly and will need to do so again if switching to exclusively to nookies.

I opened issue for us #1418

PR #1420, opened 2023-05-10
[ReferenceManager][Usability][1/n] - Collaboration Custom Lookup UI

calvinhlee23:
https://github.com/ResearchHub/researchhub-web/assets/17622930/1a62a55a-be62-43d4-a014-016d6fef5acf

### DIFF
--- a/components/ReferenceManager/form/ReferenceItemRhUserLookupInput.tsx
+++ b/components/ReferenceManager/form/ReferenceItemRhUserLookupInput.tsx
@@ -1,7 +1,13 @@
 import { isEmpty } from "~/config/utils/nullchecks";
 import { fetchUserSuggestions } from "~/components/SearchSuggestion/lib/api";
 import { ID, NullableString } from "~/config/types/root_types";
-import { ReactElement, ReactNode, SyntheticEvent, useState } from "react";
+import {
+  Fragment,
+  ReactElement,
+  ReactNode,
+  SyntheticEvent,
+  useState,
+} from "react";
 import { SuggestedUser } from "~/components/SearchSuggestion/lib/types";
 import Autocomplete from "@mui/material/Autocomplete";
 import ReferenceItemRhUserLookupInputTag from "./ReferenceItemRhUserLookupInputTag";
@@ -30,6 +36,9 @@ const getDebouncedHandleInputChange = ({
     onInputChange(queryString);
     if (!isEmpty(onLoad) && !isEmpty(queryString)) {
       onLoad(queryString);
+    } else if (isEmpty(queryString)) {
+      onFetchSucess([]);
+      debounceRef && clearTimeout(debounceRef);
     }
     debounceRef && clearTimeout(debounceRef);
     debounceRef = setTimeout(async () => {
@@ -72,6 +81,8 @@ export default function ReferenceItemRhUserLookupInput({
       fullWidth
       clearOnEscape
       clearOnBlur
+      blurOnSelect
+      selectOnFocus
       getOptionDisabled={() => true}
       getOptionLabel={
         /* Utilizing renderOption instead. Keeping this to avoid MUI errors */
@@ -80,7 +91,9 @@ export default function ReferenceItemRhUserLookupInput({
       inputValue={inputValue}
       loading={isLoading}
       loadingText={"Looking for users ..."}
-      noOptionsText={isEmpty(inputValue) ? null : "No users found"}
+      noOptionsText={
+        isEmpty(inputValue) ? "Enter user name" : "No user(s) found"
+      }
       onInputChange={handleInputChange}
       options={isEmpty(inputValue) ? [] : suggestedUsers}
       renderInput={(params) => (
@@ -103,10 +116,11 @@ export default function ReferenceItemRhUserLookupInput({
       renderOption={(_props, userOption: SuggestedUser, _state): ReactNode => {
         const { id, firstName, lastName, reputation } = userOption;
         return (
+          // Wrapping in Fragment allows proper mouse behaviors for MUI components
           <ReferenceItemRhUserLookupInputTag
             isSelectable
             onSelect={(event: SyntheticEvent): void => {
-              // event.preventDefault();
+              event.preventDefault();
               handleSelect(userOption);
             }}
             key={`RhUserLookup-Input-User-${id}-${firstName}-${lastName}-${reputation}`}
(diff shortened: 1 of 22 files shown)


components/ReferenceManager/form/ReferenceItemRhUserLookupInput.tsx
yattias, line 1: I know that this name is meaningful to you and Pat working on the project but I think it will be good to change name to one that is meaningful to rest of engineers at a glance. By looking at the name I'm led to believe it can only be used in a very specific scenario and I don't think it is the case.

Some other thoughts:
- For consistency with the other component `components/SuggestUsers`, can we just name this `SuggestUsers` or `SuggestUsersInput`?
- Since we only have one user type, we should be using `User` instead of `RhUser` in names. I realize we have a typescript type called `RhUser` but that is only because of clash in names of these two types.

I realized that you are trying to work in the confines of MUI's ecosystem and use predefined components. However, we should use our own if we already have components in place. Is there a reason for not using the `SuggestUsers` component?

components/ReferenceManager/form/ReferenceItemRhUserLookupInputTag.tsx
```diff
@@ -0,0 +1,52 @@
+import { ReactElement, SyntheticEvent } from "react";
+import { StyleSheet, css } from "aphrodite";
+import { SuggestedUser } from "~/components/SearchSuggestion/lib/types";
+import AuthorAvatar from "~/components/AuthorAvatar";
+import colors from "~/config/themes/colors";
+
+export default function ReferenceItemRhUserLookupInputTag({
```
yattias, line 7: Similar to the above, this name is quite verbose and hard to discern.

components/ReferenceManager/references/reference_organizer/ReferenceCollaboratorsSection.tsx
yattias, line 1: Are the collaborators on a project or a specific reference?

If they could be on both, can we just name `CollaboratorsSection`?

components/ReferenceManager/references/ReferencesContainer.tsx
```diff
@@ -22,6 +23,8 @@ export default function ReferencesContainer({}: Props): ReactElement {
     application: "REFERENCE_MANAGER",
     shouldRedirect: true,
   });
+  const router = useRouter();
+  const currentProjectName = router.query.project_name;
```
yattias, line 27: I realize there are some instances in our codebase where we use snake case but the vest majority is combo of snake case and kebab case. Can we update for consistence? e.g. `proejctName` instead of `project_name` ?

components/SearchSuggestion/SuggestUsers.tsx
```diff
@@ -27,22 +29,22 @@ const SuggestUsers = ({ onSelect, onChange }: Args) => {
...
+    } else {
       setUserSuggestions([]);
       setFocusedChoice(null);
     }
-    
+
     onChange(inputRef.current.textContent);
   }, [userSuggestions, inputRef?.current?.textContent]);
 
+  // NOTE: calvinhlee - this way of debouncing looks off. Need to look into this
```
yattias, line 47: I confirmed this works. You can remove the comment.

config/api.js
```diff
@@ -1,5 +1,5 @@
-// import { API } from "@quantfive/js-web-config";
-import { API } from "./api/index";
+import { API } from "@quantfive/js-web-config";
```
yattias, line 1: Probably a merge thing but I see you removed a critical line. This will cause token issues. Can you update to use `import { API } from "./api/index";` ?

yattias (changes requested):
Hey @calvinhlee23 the code is solid as usual. I did have a few comments though I think we should address before merging this. The one in `api.js` being the most critical one.

components/ReferenceManager/form/ReferenceItemRhUserLookupInput.tsx
calvinhlee23: I want to keep the name as is for now especially because it follows the design patterns that are same as other form related components in Reference Manager.

components/ReferenceManager/references/ReferencesContainer.tsx
calvinhlee23: hmm `project_name` in this case is query param in the URL. Not sure if using kebab case is a common practice in the URL

components/ReferenceManager/references/reference_organizer/ReferenceCollaboratorsSection.tsx
calvinhlee23: collaborators are participants of projects (folder). They may have influence on reference. 
The prefix of `Reference{x}` on all these file name indicattes that the component belongs to Reference as a project ( a milestone for researchhub)

components/ReferenceManager/form/ReferenceItemRhUserLookupInputTag.tsx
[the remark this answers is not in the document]
calvinhlee23: I would like to keep this in the form folder for now. I might change the folder to `shared_components` in the future.

components/ReferenceManager/form/ReferenceItemRhUserLookupInput.tsx
```diff
@@ -0,0 +1,127 @@
...
+  onUserSelect?: (user: SuggestedUser) => void;
+  placeholder?: string;
+  required?: boolean;
+  selectedUser?: SuggestedUser;
+  shouldClearOnSelect?: boolean;
+};
+
+const getDebouncedHandleInputChange = ({
+  debounceTime = 500,
+  onLoad,
```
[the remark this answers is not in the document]
calvinhlee23: Usually for calls like this, a given component needs to perform a side-effect while the call / debounce function is happening in the background. In this case, onLoad is a callback to recognize this before completing a given function of  the method. I renamed it to `onFetchLoad` to avoid confusion

components/ReferenceManager/form/ReferenceItemRhUserLookupInput.tsx
```diff
@@ -0,0 +1,127 @@
...
+    },
+  });
+
+  const handleSelect = (selectedUser: SuggestedUser): void => {
+    onUserSelect && onUserSelect(selectedUser);
+    shouldClearOnSelect && setInputValue("");
+  };
+
+  return (
+    <Autocomplete
```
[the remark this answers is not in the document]
calvinhlee23: Ah, this is great. But I'm currently utilizing MUI's options component. Which i don't think recognizes key strokes to focus-toggle on options. If we think this should be implemented, I'll keep this in mind and make a custom component all together.

PR #1421, opened 2023-05-10
Multipdf upload

lightninglu10:
https://www.loom.com/share/211a2cac968745bba23cef1a6e17c4a2

Allows users to drag & drop PDF's into the citation manager

Each dragged & dropped PDF will load in when the processing is done on it

### DIFF
--- a/components/ReferenceManager/references/ReferencesContainer.tsx
+++ b/components/ReferenceManager/references/ReferencesContainer.tsx
@@ -182,10 +182,7 @@ function ReferencesContainer({
                       }}
                       endAdornment={
                         <InputAdornment position="end">
-                          <IconButton
-                            aria-label="toggle password visibility"
-                            edge="end"
-                          >
+                          <IconButton edge="end">
                             <i
                               className="fa-regular fa-magnifying-glass"
                               style={{ fontSize: 16 }}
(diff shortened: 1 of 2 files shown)


components/DroppableZone.tsx
yattias, line 1: I like this component. It is well structured, named and `Props` are easy to understand. Nice!

components/DroppableZone.tsx
```diff
@@ -0,0 +1,108 @@
...
+    alignItems: "center",
+    width: "100%",
+    height: "100%",
+    boxSizing: "border-box",
+    padding: "20px 0px",
+    transition: "all ease-out 0.1s",
+  },
+  dropzoneContainer: {
+    boxSizing: "border-box",
+    width: "calc(100% - 240px)",
```
yattias, line 78: Would this line prevent the component from being used elsewhere? Otherwise, seems like a pretty reusable solid component.

components/ReferenceManager/references/ReferencesContainer.tsx
```diff
@@ -51,85 +112,104 @@ export default function ReferencesContainer({}: Props): ReactElement {
...
+                          border: "0px !important",
+                          "&:hover": {
+                            border: "0px",
+                          },
+                        },
+                      }}
+                      endAdornment={
+                        <InputAdornment position="end">
+                          <IconButton
+                            aria-label="toggle password visibility"
```
yattias, line 186: Probably an icon thing should be remove?

yattias (approved):

PR #1427, opened 2023-05-16
improve HMR build time

victor-perkins226:


yattias:
@victor-perkins226 can you include some banchmarks for before / after so we can compare the HMR speed difference?

victor-perkins226:
@yattias Should I add benchmark on all pages or components?

victor-perkins226:
I improved HMR by modifying next config.
I will add benchmark on each page, so that you can check loading time.

yattias:
@victor-perkins226 Could be a video, description or screenshot of before and after speed improvements

victor-perkins226:
That's good idea

victor-perkins226:
Added benchmark and video

PR #1448, opened 2023-06-05
Anonymity toggle

Mozes721:


components/Comment/CommentAnonymously.tsx
```diff
@@ -0,0 +1,29 @@
+import { ReactElement } from "react";
+import Switch from "@mui/material/Switch";
+import { css, StyleSheet } from "aphrodite";
+
+type Props = {
+  id: string;
```
calvinhlee23, line 6: We have a root type called ID! :) 
```suggestion
  id: ID;
```

components/Comment/CommentAnonymously.tsx
```diff
@@ -0,0 +1,29 @@
...
+  anonymous: boolean;
```
calvinhlee23, line 7: just a nit code practice stuff that we try to start flags with verbs
```suggestion
  isAnonymous: boolean;
```

components/Comment/CommentAnonymously.tsx
```diff
@@ -0,0 +1,29 @@
...
+  onToggle: () => void;
+};
+
+const AnonymousToggle = ({ id, anonymous, onToggle }: Props): ReactElement => {
+  return (
+    <div className={css(styles.toggleContainer)}>
+      toggle anonymity
```
calvinhlee23, line 14: typing these html strings as interpolated string makes it easy to code-search for other developers. Also, it keeps the trailing / leading white spaces as typed rather than relying on the browser to interpret these themselves ( usually browsers remove them ) 
```suggestion
      {"toggle anonymity"}
```

components/Comment/CommentEditor.tsx
```diff
@@ -67,6 +68,7 @@ const CommentEditor = ({
   handleClose,
   editorStyleOverride,
 }: CommentEditorArgs) => {
+  const [anonymous, setAnonymous] = useState<boolean>(true);
```
calvinhlee23, line 71: See comment above
```suggestion
  const [isAnonymous, setIsAnonymous] = useState<boolean>(true);
```

components/Comment/lib/api.ts
```diff
@@ -1,16 +1,18 @@
-import {
-  ID,
-  RHUser,
-  RhDocumentType,
-  parseUser,
-} from "~/config/types/root_types";
+import { ID, RhDocumentType } from "~/config/types/root_types";
```
calvinhlee23, line 1: Please try to keep the imports and arguments in alphabethical order. If you use VS code, you can highlight the lines and press Fn+F9, it alphabetizes for you

components/Comment/lib/api.ts
```diff
@@ -1,16 +1,18 @@
...
 import { Helpers } from "@quantfive/js-web-config";
 import config from "./config";
 import { sortOpts } from "./options";
 import { parseVote, Vote } from "~/config/types/vote";
 import uniqBy from "lodash/uniqBy";
+import { buildApiUri } from "~/config/utils/buildApiUri";
+
+type AnonArgs = {
+  id: string;
+  // eslint-disable-next-line no-unused-vars
```
calvinhlee23, line 13: what is this :o

calvinhlee23 (changes requested):

components/Comment/CommentAnonymously.tsx
Mozes721: Adjusted

components/Comment/CommentAnonymously.tsx
Mozes721: refactored

PR #1457, opened 2023-06-13
[RefManager] - Alpha Release Version

calvinhlee23:
### BE: 
- depends on https://github.com/ResearchHub/researchhub-backend/pull/1294

### Screenshot: 
Screenshot 2023-06-13 at 11 40 17 AM
Screenshot 2023-06-13 at 11 40 20 AM
Screenshot 2023-06-13 at 11 40 26 AM

### DIFF
--- a/components/ReferenceManager/references/ReferencesContainer.tsx
+++ b/components/ReferenceManager/references/ReferencesContainer.tsx
@@ -405,14 +405,9 @@ function ReferencesContainer({
                     marginBottom: "20px",
                   }}
                 >
-                  <div
-                    style={
-                      isEmpty(selectedReferenceIDs)
-                        ? { visibility: "hidden" }
-                        : undefined
-                    }
-                  >
+                  <div style={{ display: "flex", alignItems: "center" }}>
                     <DropdownMenu
+                      disabled={isEmpty(selectedReferenceIDs)}
                       menuItemProps={[
                         {
                           itemLabel: `Export reference${
@@ -458,6 +453,7 @@ function ReferencesContainer({
                       }
                       size="medium"
                     />
+                    <Typography sx={{marginLeft: "8px"}}>{`${selectedReferenceIDs.length} selected`}</Typography>
                   </div>
                   <div
                     className="ReferenceContainerSearchFieldWrap"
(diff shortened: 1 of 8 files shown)


components/ReferenceManager/references/ReferencesContainer.tsx
```diff
@@ -232,6 +231,23 @@ function ReferencesContainer({
     setLoading(false);
   };
 
+  useEffectFetchOrgProjects({
```
yattias, line 234: The functionality of these is good but as far as naming goes, I think we can omit the "Effect" and "Fetch" portions and just name: `useOrgProjects`.

This is common pattern in other packages and removes the need to know that the hook uses a "useEffect" or "fetch" behind the scenes. From the component using it's perspective, all that is needed to know is that the hook can be used to get projects somehow. 

Not a release blocker, just a suggestion.

components/ReferenceManager/references/reference_table/ReferencesTable.tsx
```diff
@@ -131,17 +131,36 @@ export default function ReferencesTable({
...
+                  (item) => item.id === params?.row?.id
+                )
+              ),
+            });
             setIsDrawerOpen(true);
+          } else {
+            setSelectedReferenceIDs(params.id);
           }
         }}
+        // onCellDoubleClick={}
```
yattias, line 149: Probably remove this?

config/api.js
```diff
@@ -7,7 +7,8 @@ import { convertToBackendFilters } from "~/components/UnifiedDocFeed/utils/conve
 export const apiRoot = {
   production: "backend.researchhub.com",
   staging: "staging-backend.researchhub.com",
-  dev: "localhost:8000",
+  // dev: "localhost:8000",
+  dev: "127.0.0.1:8000" /* localhost */,
```
yattias, line 11: Can we keep localhost? There is some functionality that utilizes "localhost" as the HOST

yattias (approved):

components/ReferenceManager/references/ReferencesContainer.tsx
calvinhlee23: Will keep that in mind. Thanks for the tip!

PR #1460, opened 2023-06-14
Edit metadata, JSON+LD

yattias:
- Adding ability to edit paper metadata / abstract
- Fixing JSON+LD structured data issues

**Screenshot 1**
image

**Screenshot 2**
image

**Screenshot 3**
image

### DIFF
--- a/components/Document/DocumentHeaderV2.tsx
+++ b/components/Document/DocumentHeaderV2.tsx
@@ -14,7 +14,6 @@ import {
   Paper,
   isPaper,
   isPost,
-  parsePaper,
 } from "./lib/types";
 import DocumentVote from "./DocumentVote";
 import PermissionNotificationWrapper from "../PermissionNotificationWrapper";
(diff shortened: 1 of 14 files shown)


components/Document/DocumentHeaderV2.tsx
```diff
@@ -86,18 +87,42 @@ const DocumentHeader = ({ document: doc, metadata }: Props) => {
...
-    //         label: "Edit metadata",
-    //         icon: <FontAwesomeIcon icon={faPen} />,
-    //         value: "edit-metadata",
-    //         onClick: () => {
-    //           alert("implement me");
-    //         },
-    //       },
-    //     ]
-    //   : []),
+    ...(isPaper(doc)
```
calvinhlee23, line 90: Hmm.. doing checks like this looks already off. I just looked at what the function is doing, and it looks like it's just type-checking. 
If we are going to consolidate different documents to be viewed with a single component like this (which I like) shouldn't we have a consolidate payload from the backend. I believe BE already has to do something like this before it sends data to FE. 
In that case, might as well just put off this cumbersome parsing all off to BE instead of parsing twice in BE & FE.

calvinhlee23 (changes requested):

calvinhlee23:
The modal looks good. 
Had some thoughts on the DocumentView

components/Document/DocumentHeaderV2.tsx
yattias: I'll share some thought.

We have `Paper` and `Post` types which are a little different from each other. Each of these extends `GenericDocument` so essentially a Post is a document and also Paper.

A paper for instance is defined as such:

```
export type Paper = GenericDocument & {
  journal?: string;
  isOpenAccess: boolean;
  laymanTitle: string;
  externalUrl?: string;
  abstract?: string;
  abstractHtml?: TrustedHTML;
  license?: string;
};
```

In the above case, since we don't know what type document is, we need to use `isPaper` which makes an assertion and avoids TS errors.

```
export const isPaper = (document: GenericDocument): document is Paper => {
  return (document as Paper).type === "paper";
};
```

The main takeaway above is that the return value is `document is Paper`

PR #1531, opened 2023-07-25
codespell: add config and action to codespell the code to avoid known typos

yarikoptic:
Some changes are to the code but upon quick look should be good unless you know them somehow being used outside.  Let me know - I could exclude them (requester etc)

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,0 +1,4 @@
+[codespell]
+skip = .git,*.pdf,*.svg,package-lock.json
+#
+# ignore-words-list =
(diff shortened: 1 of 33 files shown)


yattias:
Thanks @yarikoptic for submitting this PR. How can we run codespell on a go-forward basis?

yarikoptic:
well -- github action would run it for you.  To run locally, just `pip install codespell` and then `codespell`.  You could also add it to `pre-commit` if you start using that one, see e.g. config in https://github.com/incf-nidash/PyNIDM/blob/master/.pre-commit-config.yaml .

PR #1555, opened 2023-08-15
Reference Manager Search

lightninglu10:
Integrating search into reference manager

### DIFF
--- a/components/ReferenceManager/references/ReferencesContainer.tsx
+++ b/components/ReferenceManager/references/ReferencesContainer.tsx
@@ -6,6 +6,7 @@ import {
 import {
   faArrowUpFromBracket,
   faFolderPlus,
+  faMagnifyingGlass,
   faPlus,
   faTrashXmark,
 } from "@fortawesome/pro-light-svg-icons";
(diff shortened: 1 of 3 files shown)


components/ReferenceManager/references/ReferencesContainer.tsx
```diff
@@ -152,6 +162,53 @@ function ReferencesContainer({
     ReactTooltip.rebuild();
   }, [selectedReferenceIDs]);
 
+  const { setReferenceTableRowData } = useReferencesTableContext();
+
+  const fetchCitationsWithQuery = async (url) => {
+    const config = api.GET_CONFIG();
+    config.headers["X-organization-id"] = currentOrg?.id.toString();
```
yattias, line 169: In my branch I updated the different config such as `GET_CONFIG` to include optional headers. This is okay too for now.

yattias (approved):

components/ReferenceManager/references/ReferencesContainer.tsx
lightninglu10: Ah nice. Will keep this for now and then we can make it more explicit in your branch

PR #1565, opened 2023-08-23
Row selection-related reference manager fixes

yattias:
At a high level, configured row selection mechanism to work like Dropbox and Google Drive for desktop and touch-enabled devices.

- [x] - Single click anywhere in the row selects the whole row
- [x] - Double click on folder to open folder
- [x] - Double click on reference to open viewer
- [x] - Make "Go back to parent" not selectable
- [x] - Disable click actions if row is loading
- [x] - Single click on touch-capable devices triggers desktop double click action
- [x] - Configure meta key selection to allow for multiple selected rows
- [x] - Preventing row actions from being cut off in smaller screens
- [x] - Single state for selected rows instead of two state vars (folders, refs). Two state vars interferes with some DataGrid attributes
- [x] - Show actions by default for touch capable devices
- [x] - Allow multi-row selection on click if touch is enabled
- [x] - Show "more options" (three dots) for touch devices by default
- [x] - Do not show checkbox if row is loading
- [x] - Wire edit metadata, delete to "more options" menu

### DIFF
--- a/components/ReferenceManager/references/reference_table/ReferencesTable.tsx
+++ b/components/ReferenceManager/references/reference_table/ReferencesTable.tsx
@@ -46,6 +46,7 @@ import OpenWithOutlinedIcon from "@mui/icons-material/OpenWithOutlined";
 import ReferenceItemOptsDropdown from "../reference_item/ReferenceItemOptsDropdown";
 import { useEffectHandleClick } from "~/config/utils/clickEvent";
 import { faFolderOpen } from "@fortawesome/pro-solid-svg-icons";
+import { useHasTouchCapability } from "~/config/utils/device";
 
 type Props = {
   createdReferences: any[];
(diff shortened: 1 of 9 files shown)


components/ReferenceManager/references/reference_table/ReferencesTable.tsx
```diff
@@ -246,6 +273,61 @@ export default function ReferencesTable({
...
+      // If meta key is pressed, add or remove from the selection
+      if (selectedRows.includes(params.id)) {
+        handleRowSelection(selectedRows.filter((id) => id !== params.id));
+      } else {
+        handleRowSelection([...selectedRows, params.id]);
+      }
+    } else {
+      // If meta key is not pressed, replace the selection
+      // Without timeout, double click will not be reached since the state update will interrupt it.
+      setTimeout(() => {
```
lightninglu10, line 293: I always get nervous about putting setTimeout's for rendering purposes. Some devices may take longer than 150ms and then it won't work and we have phantom bugs. 

Is there anyway to build without this?

yattias: Yeah, I understand the nervousness around setTimeout. It will have no impact on rendering. The reason it is needed   is  because we are  using both single  and double click events.  The setTimeout  is recommendation from various sources.  I tried to do without it  -  and it is possible, but the UX suffers a bit

PR #1567, opened 2023-08-27
Responsive layout for Ref manager

yattias:


components/ReferenceManager/references/ReferencesContainer.tsx
```diff
@@ -194,6 +207,39 @@ function ReferencesContainer({
     }
   };
 
+  const addFolderToChildren = (result) => {
```
yattias, line 210: These functions were moved from `BasicToggleableNavbarLeft`

components/Form/Button.js
```diff
@@ -121,6 +121,7 @@ const styles = StyleSheet.create({
     textDecoration: "none",
     fontWeight: 500,
     lineHeight: 1.75,
+    fontFamily: "Roboto, sans-serif",
```
lightninglu10, line 124: are our default fonts not Roboto?

components/ReferenceManager/references/reference_uploader/ReferenceManualUploadDrawer.tsx
```diff
@@ -80,6 +82,14 @@ export default function ReferenceManualUploadDrawer(): ReactElement {
     setReferenceSchemaValueSet,
   });
 
+  useEffect(() => {
+    const rootLeftSidebarWidth =
+      document.querySelector(".root-left-sidebar")?.clientWidth || 0;
+    const refManagerNavbarWidth =
+      document.querySelector(".ToggleableNavbarLeft")?.clientWidth || 0;
+    setCalculatedLeftMargin(rootLeftSidebarWidth + refManagerNavbarWidth + 2);
```
lightninglu10, line 90: Will this cause any flashing? If we server side render some pieces of this, and then grab things from client side, typically the app will flash

lightninglu10: Is it better here to set a constant width for some of these classes so that we don't have to set this in javascript?

components/contexts/NavigationContext.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const [isRefManagerDisplayedAsDrawer, setIsRefManagerDisplayedAsDrawer] =
+    useState(false);
+
+  const { width: winWidth, height: winHeight } = useWindow();
+
+  useEffect(() => {
+    const displayAsDrawer = winWidth && winWidth < breakpoints.medium.int;
+
+    if (displayAsDrawer && !isRefManagerDisplayedAsDrawer) {
+      setIsRefManagerDisplayedAsDrawer(true);
```
lightninglu10, line 36: Typically I like handling this in CSS instead of JS if we can. 

Having CSS depend on JS can lead to bugs / flashing issues.

Is it necessary to handle these widths here like this?

components/Form/Button.js
yattias: I realized that our Button component doesn't inherit the font family. Kinda unexpected, but for some reason the default browser stylesheet overrides our button font family. Confirmed on production.

components/ReferenceManager/references/reference_uploader/ReferenceManualUploadDrawer.tsx
yattias: It should not cause flashing because it executes within a `useEffect` which only fires on the client side.

Initially, these values were hard coded but I couldn't rely on the hard-coded component widths because some components will not be visible in certain resolutions. For example, the left-most sidebar is not visible in mobile screens. The ref. manager sidebar turns into a drawer at 900 px, etc...

Needed a more accurate solution.

PR #1580, opened 2023-09-27
Upload pdf from doc viewer

lightninglu10:
If PDF url doesn't exist, allow user to upload pdf from doc viewer

Screenshot 2023-09-27 at 3 24 45 PM

### DIFF
--- a/components/Document/DocumentViewer.tsx
+++ b/components/Document/DocumentViewer.tsx
@@ -14,7 +14,10 @@ import throttle from "lodash/throttle";
 import { LEFT_SIDEBAR_MAX_WIDTH } from "../Home/sidebar/RootLeftSidebar";
 import DocumentPlaceholder from "./lib/Placeholders/DocumentPlaceholder";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
-import { faCircleExclamation } from "@fortawesome/pro-light-svg-icons";
+import {
+  faCircleExclamation,
+  faSpinnerThird,
+} from "@fortawesome/pro-light-svg-icons";
 import {
   Comment as CommentType,
   CommentThreadGroup,
@@ -28,6 +31,13 @@ import DocumentViewerContext, {
 } from "./lib/DocumentViewerContext";
 import { useRouter } from "next/router";
 import colors from "~/config/themes/colors";
+import UploadFileDragAndDrop from "../UploadFileDragAndDrop";
+import postUploadFiles from "../ReferenceManager/references/api/postUploadFiles";
+import { nullthrows } from "~/config/utils/nullchecks";
+import { useReferenceActiveProjectContext } from "../ReferenceManager/references/reference_organizer/context/ReferenceActiveProjectContext";
+import { useOrgs } from "../contexts/OrganizationContext";
+import { getCurrentUser } from "~/config/utils/getCurrentUser";
+import { updateReferenceCitationFile } from "../ReferenceManager/references/api/updateReferenceCitation";
 
 const AnnotationLayer = dynamic(
   () => import("~/components/Comment/modules/annotation/AnnotationLayer")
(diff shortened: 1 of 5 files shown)


components/Document/DocumentViewer.tsx
```diff
@@ -300,6 +322,43 @@ const DocumentViewer = ({
   const shouldScroll =
     actualContentWidth > windowDimensions.width - LEFT_SIDEBAR_MAX_WIDTH;
 
+  const uploadPDF = (acceptedFiles: File[] | any[]): void => {
+    // const preload: Array<PreloadRow> = [];
```
yattias, line 326: Can we remove this?

components/Document/DocumentViewer.tsx
```diff
@@ -300,6 +322,43 @@ const DocumentViewer = ({
   const shouldScroll =
     actualContentWidth > windowDimensions.width - LEFT_SIDEBAR_MAX_WIDTH;
 
+  const uploadPDF = (acceptedFiles: File[] | any[]): void => {
```
yattias, line 325: The function itself is sound. Ideally, I'd like to have it one level up (in whatever is referencing DocumentViewer) so that `DocumentViewer` isn't aware or concerned with references. That way we can it more pure and also reduce the number of props, mainly removing:

  setReferenceItemDatum,
  referenceItemDatum,

components/Document/DocumentViewer.tsx
```diff
@@ -54,14 +64,18 @@ type Props = {
...
+  setReferenceItemDatum?: (datum) => void;
+  referenceItemDatum: any;
 };
 
 const DocumentViewer = ({
   postHtml,
   onZoom,
   viewerWidth = config.width,
   pdfUrl,
+  isPost,
```
yattias, line 78: Instead of including this prop, you can just use the `doc` being passed in.

`Doc` is defined as a `GenericDocument` but you can use a typescript helper function called `isPost` to identify the type of object. That way, we can reduce the number of props. It is defined in `Document/lib/types`

components/Document/DocumentViewer.tsx
```diff
@@ -364,24 +424,62 @@ const DocumentViewer = ({
...
+                              spin
+                              style={{ fontSize: 30 }}
+                            />
+                          ) : (
+                            ""
+                          )
+                        }
+                        accept={".pdf"}
+                        beforeImageInstructions={
+                          <h3 className={css(styles.uploadFileTitle)}>
```
yattias, line 465: I think we should tell them that we don't have the PDF since the user might expect us to have it. Maybe we put some caption saying "We couldn't automatically fetch the PDF"

components/Document/DocumentViewer.tsx
```diff
@@ -364,24 +424,62 @@ const DocumentViewer = ({
...
+                            Upload your pdf
+                          </h3>
+                        }
+                        // extraInstructions={
```
yattias, line 469: Can we remove

components/Document/DocumentViewer.tsx
```diff
@@ -364,24 +424,62 @@ const DocumentViewer = ({
...
+                        //   <div className={css(styles.extraInstructions)}>
+                        //     We weren't able to fetch the PDF
+                        //   </div>
+                        // }
+                        handleFileDrop={
+                          uploadPDF
+                          // formattedReferenceRows.length === 0
```
yattias, line 476: Can we remove

components/Document/DocumentViewer.tsx
```diff
@@ -300,6 +322,43 @@ const DocumentViewer = ({
...
+    updateReferenceCitationFile({
+      payload: {
+        // TODO: calvinhlee - create utily functions to format these
+        citation_id: referenceItemDatum.id,
+        organization: currentOrg.id,
+        attachment: acceptedFiles[0],
+      },
+      onError: (e): void => {
+        console.log(e);
+        setUploadingPdf(false);
```
yattias, line 349: On Error, can we put up an `alert("Upload failed. Try again later")` or something like that? It's an uncommon scenario but would be good to have the user feedback

PR #1581, opened 2023-09-28
Get tweet count for Biorxiv

lightninglu10:
Showing tweet count in bottom right of the document header. Also refreshes the tweet score for every visit to this page.

Screenshot 2023-09-27 at 5 03 31 PM

### DIFF
--- a/components/Document/DocumentHeaderV2.tsx
+++ b/components/Document/DocumentHeaderV2.tsx
@@ -59,6 +59,22 @@ const DocumentHeader = ({ document: doc, metadata, twitterScore }: Props) => {
     isEmpty(state.auth?.user) ? null : parseUser(state.auth.user)
   );
 
+  const twitterScoreRender = (props) => {
+    const { className } = props;
+    return (
+      <div
+        className={css(styles.twitterScore, className)}
+        data-tip={`This paper has been tweeted ${twitterScore} time${
+          twitterScore && twitterScore > 1 ? "s" : ""
+        }`}
+        data-for={"twitter-score-tooltip"}
+      >
+        <FontAwesomeIcon icon={faTwitter} /> {twitterScore}
+        <ReactTooltip effect="solid" id="twitter-score-tooltip" place="top" />
+      </div>
+    );
+  };
+
   const tabs = getTabs({
     router,
     document: doc,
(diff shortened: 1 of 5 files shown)


components/Document/DocumentHeaderV2.tsx
```diff
@@ -271,6 +299,26 @@ const styles = StyleSheet.create({
...
+    alignItems: "center",
+    border: `1px solid ${colors.GREY_BORDER}`,
+    color: colors.MEDIUM_GREY(),
+    padding: "6px 12px",
+    borderRadius: 50,
+    fontSize: 14,
+    marginLeft: "auto",
+    height: "fit-content",
+    cursor: "default",
+    // cursor: "pointer",
```
yattias, line 315: Can we remove?

components/Document/lib/sharedGetStaticProps.ts
```diff
@@ -30,6 +30,7 @@ export default async function sharedGetStaticProps({
 
   try {
     documentData = await fetchDocumentByType({ documentType, documentId });
+    console.log(documentData);
```
yattias, line 33: Can we remove?

pages/paper/[documentId]/[documentSlug]/index.tsx
```diff
@@ -60,6 +65,21 @@ const DocumentIndexPage: NextPage<Args> = ({
   }) as [Paper | null, Function];
   const { revalidateDocument } = useCacheControl();
 
+  const refetchTwitterScore = async () => {
```
Comment (yattias) [inline pages/paper/[documentId]/[documentSlug]/index.tsx:68]: Are you refetching this value here because you want to make sure we get "fresh" data? If so, I think the most consistent way to handle this would be to add it to the metadata endpoint `get_document_metadata`. I realize we chatted and you had a bunch of issues related. I can take a look if you are unable to get it to work

pages/paper/[documentId]/[documentSlug]/index.tsx
```diff
@@ -93,6 +113,7 @@ const DocumentIndexPage: NextPage<Args> = ({
         errorCode={errorCode}
         metadata={documentMetadata}
         documentType={documentType}
+        twitterScore={twitterScore}
```
Comment (yattias) [inline pages/paper/[documentId]/[documentSlug]/index.tsx:116]: I think `twitterScore` should live within `metadata`. We can chat about it if it isn't working

yattias (changes requested):

PR #1582, opened 2023-09-28
Hubs V2

yattias:


components/Document/DocumentViewer.tsx
lightninglu10, line 1: If we remove this minHeight, will the doc viewer be fine in the ref manager too?

components/Hubs/HubTag.tsx
```diff
@@ -0,0 +1,48 @@
...
+import IconButton from "~/components/Icons/IconButton";
+import { StyleSheet, css } from "aphrodite";
+import colors from "~/config/themes/colors";
+
+const HubTag = ({ hub }: { hub: Hub }) => {
+  return (
+    <Link
+      key={`/hubs/${hub.slug ?? ""}-index`}
+      href={`/hubs/${hub.slug}`}
+      onClick={(e) => e.stopPropagation()}
```
lightninglu10, line 13: Do we need a stopPropagation here? Sometimes this will lead to bugs. Shouldn’t the link should take care of itself without the need for a stopPropagation?

components/Hubs/api/fetchHubs.js
```diff
@@ -2,20 +2,14 @@ import API from "~/config/api";
 import { Helpers } from "@quantfive/js-web-config";
 import { emptyFncWithMsg } from "~/config/utils/nullchecks";
 
-export const getHubs = () => {
+export const getHubs = async ({ ordering = "score" }) => {
```
lightninglu10, line 5: We don’t need the async decorator here if we’re not using await in the function

pages/hubs/api/fetchHubFromSlug.ts
```diff
@@ -7,6 +7,7 @@ export default function fetchHubFromSlug({ slug }: { slug: string }): any {
     .then(Helpers.checkStatus)
     .then(Helpers.parseJSON)
     .then((res: any): any => {
+      console.log("res?.results", res?.results);
```
lightninglu10, line 10: Remove

lightninglu10 (commented):
Some small feedback, not fully done reviewing everything but it looks mostly fine

PR #1586, opened 2023-10-04
Replication Prediction Market (No RSC)

manveerxyz:
Implement Replication Prediction Markets, without RSC voting (RSC voting is implemented in [manveer/v0-repl-market-rsc](https://github.com/manveerxyz/researchhub-web/tree/manveer/v0-repl-market-rsc))

[Design File](https://www.figma.com/file/5nYfZkYPjpDQz9cFLeMUVV/Research-Hub?type=design&node-id=18321%3A7355&mode=design&t=OLSiEb9pB0WwmYEL-1).

### Technical Details
- Fetches prediction market details (number of votes/yes/no) in `getSharedStaticProps`
- Fetches user previous vote and list of all votes after page load using `useEffect`
- Prevents voting if paper's author ID matches current user's author ID

### Demo

https://github.com/ResearchHub/researchhub-web/assets/16143968/50a8e8e6-c3ab-4ab6-8b42-6bb9e5d4171d
- Requires login
- Creates vote on backend and updates local web state
- User can update vote and delete vote
- Reloads state and user's vote
- Lists votes at bottom
- Shows votes on author page
- Add comments on replicability

### Demo with RSC
This isn't implemented in this PR, but it's just for reference.
[Design File for RSC Voting](https://www.figma.com/file/5nYfZkYPjpDQz9cFLeMUVV/Research-Hub?type=design&node-id=18436%3A10778&mode=design&t=D5QTAg4hpOy7C2Fq-1).
<details>
<summary>Demo Video</summary>
<br>

https://github.com/ResearchHub/researchhub-web/assets/16143968/6786046e-e7a8-4414-8a5c-62f4e9b275e6

</details>

### Mobile Design

| No Votes | Yes Vote |
|---------|---------|
| Screenshot 2023-10-07 at 9 47 41 AM | Screenshot 2023-10-07 at 9 47 15 AM |


components/Comment/CommentEditor.tsx
```diff
@@ -147,7 +147,9 @@ const CommentEditor = ({
       if (_isEmpty) {
         forceShowPlaceholder({
           quillRef,
-          placeholderText: placeholder,
+          placeholderText:
+            commentTypes.find((ctype) => ctype.value === commentType)
+              ?.placeholder || placeholder,
```
manveerxyz, line 152: The type-specific placeholder weren't showing because they were getting overwritten by the default `placeholder`. This fixes that issue.

components/Document/lib/sharedGetStaticProps.ts
```diff
@@ -84,13 +84,15 @@ export default async function sharedGetStaticProps({
...
       const filter = getCommentFilterByTab(tabName);
-      commentData = await fetchCommentsAPI({
-        documentId,
-        documentType:
-          documentType === "post" ? "researchhubpost" : documentType,
-        // @ts-ignore
-        filter,
-      });
+      if (filter !== null) {
+        commentData = await fetchCommentsAPI({
```
lightninglu10, line 88: This might be out of sync with master now, I believe we've moved this out of the getStaticProps (was causing a bug due to too much data)

components/Document/pages/DocumentReplicationMarketPage.tsx
```diff
@@ -0,0 +1,351 @@
+import { NextPage } from "next";
```
lightninglu10, line 1: Is this page needed? We may be able to host the page by itself in the `pages/paper/[documentId]/[documentSlug]/[tabName]` 

That would keep our pages consistent

components/Form/Dropdown.tsx
```diff
@@ -0,0 +1,153 @@
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
```
lightninglu10, line 1: What is this dropdown form for? I believe we already have a dropdown component. OK to add another, just need to understand the use case

components/Document/lib/sharedGetStaticProps.ts
yattias: That is correct

components/Form/Dropdown.tsx
yattias: Right. We have `GenericMenu` which is just a generic dropdown used in a bunch of places. We also have a few others I believe but that one is pretty light weight

yattias (commented):
Good stuff here. I appreciate that you used patterns and component that we have in place instead of making your own. It is very obvious you spent time studying the codebase.

components/Document/pages/DocumentReplicationMarketPage.tsx
manveerxyz: I tried creating a separate page at `pages/paper/[documentId]/[documentSlug]/replicability.tsx` but couldn't get it to work — something about the `tabName` param or how the page was structured was interfering with some logic I think.

It's my first time working with Next so wasn't able to figure out the mechanics of why.

I can try to refactor if we think that's the best move, but based off this thread it seems potentially okay to leave as-is?

components/Form/Dropdown.tsx
manveerxyz: Ah yeah this was leftover from the branch with RSC — removed the component.

config/api.js
```diff
@@ -1306,6 +1306,44 @@ const routes = (BASE_URL) => {
         );
       }
     },
+
+    PREDICTION_MARKET: ({ marketId } = {}) => {
```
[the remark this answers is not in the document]
manveerxyz: Fair point, moved it into `components/PredictionMarket/api`

components/Document/pages/DocumentReplicationMarketPage.tsx
lightninglu10: The way the routes work here is that [tabName] is a dynamic parameter. IMO you should put all of your Replication code into a component and then render it here based on the tabName, not create a separate page for it.

lightninglu10: So you could have an if statement, and do like this: 

```
import { GetStaticPaths, GetStaticProps, NextPage } from "next";
import DocumentCommentsPage from "~/components/Document/pages/DocumentCommentsPage";
import sharedGetStaticProps from "~/components/Document/lib/sharedGetStaticProps";

interface Args {
  documentData?: any;
  commentData?: any;
  errorCode?: number;
  tabName: string;
  metadata?: any;
}

const TabPage: NextPage<Args> = ({
  documentData,
  commentData,
  tabName,
  errorCode,
  metadata,
}) => {
  if (tabName === 'replicability') {
    return (<PaperReplication />)
  }
  return (
    <DocumentCommentsPage
      documentData={documentData}
      documentType="paper"
      metadata={metadata}
      tabName={tabName}
      commentData={commentData}
      errorCode={errorCode}
    />
  );
};

export const getStaticProps: GetStaticProps = async (ctx) => {
  return sharedGetStaticProps({ ctx, documentType: "paper" });
};

export const getStaticPaths: GetStaticPaths = async () => {
  return {
    paths: [
      //   {
      //   params: {
      //     documentType: 'paper',
      //     documentId: '1276082',
      //     documentSlug: 'boundary-vector-cells-in-the-goldfish-central-telencephalon-encode-spatial-information',
      //     tabName: 'conversation',
      //   },
      // },
    ],
    fallback: true,
  };
};

export default TabPage;
```

manveerxyz: Isn't that what I did? Ref: https://github.com/ResearchHub/researchhub-web/pull/1586/files#diff-bc02f4ca9eeb4dfc8a2faac8b55e35515b8b25e298300833319d6709f2cfcd1dR20

PR #1593, opened 2023-10-17
Answer bounty

lightninglu10:
Screenshot 2023-10-16 at 1 31 35 PM

Making the bounty an answer CTA not contribute

### DIFF
--- a/components/Comment/CommentActions.tsx
+++ b/components/Comment/CommentActions.tsx
@@ -310,7 +310,7 @@ const CommentActions = ({ comment, toggleReply, onBountyAdd }: Args) => {
             <span className={css(styles.actionText)}>Copy link</span>
           </IconButton>
         </div>
-        {comment.bounties[0] && (
+        {comment.bounties[0] && comment.bounties[0]?.status !== "CLOSED" && (
           <div
             className={`${css(styles.action, styles.copyLinkAction)} link-btn`}
             data-tip={tooltipText}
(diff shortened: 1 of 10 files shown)


yattias:
@lightninglu10 Can you hide the contribute for inline comments? 

image


You can do this:

```
{![COMMENT_CONTEXTS.ANNOTATION, COMMENT_CONTEXTS.REF_MANAGER].includes(
          context
        ) && (
          <div
            className={`${css(styles.action, styles.actionReply)} reply-btn`}
          >
              /* Show button*/
          </div>
        )}
        ```

File: components/Comment/CommentActions.tsx [hunk 1]
```diff
@@ -304,6 +310,22 @@ const CommentActions = ({ comment, toggleReply }: Args) => {
             <span className={css(styles.actionText)}>Copy link</span>
           </IconButton>
         </div>
+        <div
```
yattias, line 313: The hover effect is missing from this button

yattias: See how "Copy Link" and "Reply" have a hover effect

yattias:
@lightninglu10 Can you remove "Summary" as an option? Also, (and this may already be implemented after BE is deployed), could we default to "Peer Review" ?

image

yattias:
I think the call to create a review isn't being fired when creating a peer review. 
There are two calls that should fire: One for creating a peer review and another for creating a comment

image

yattias:
@lightninglu10 Can we remove "Contribute to bounty" from replies to bounty?

image

yattias:
I created a couple of peer reviews as replies to a bounty but the peer reviews do not show up in the "Peer review" tab

PR #1595, opened 2023-10-18
Bounty types

lightninglu10:
Adding bounty types to bounty creation modal

Screenshot 2023-10-18 at 12 08 45 PM

### DIFF
--- a/components/Bounty/BountyModal.tsx
+++ b/components/Bounty/BountyModal.tsx
@@ -28,6 +28,8 @@ import ContentBadge from "../ContentBadge";
 import { isEmpty } from "~/config/utils/nullchecks";
 import { RootState } from "~/redux";
 import { ID, parseUser } from "~/config/types/root_types";
+import FormSelect from "../Form/FormSelect";
+import { COMMENT_TYPE_OPTIONS } from "../Comment/lib/types";
 
 type Props = {
   isOpen: boolean;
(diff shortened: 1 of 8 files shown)


yattias (commented):
@lightninglu10 Code looks good but I think the modal needs minor tweaking.

1. The subheader and title font treatment is too close to each other. Looks like two headers. Can we make subheader a bit smaller and remove bold?
2. I don't think "Discussion" bounty type doesn't make sense from a user perspective. Maybe we keep "Discussion" in the backend but have "Other" as the option?
3. Can we change "Answer" to something like "Answer to Question" ?

Here is a screenshot from stackoverflow

image

lightninglu10:
I think the header "Add Bounty" and the other text are OK. I tweaked it slightly, but here are designs for the feature: https://www.figma.com/file/5nYfZkYPjpDQz9cFLeMUVV/Research-Hub?node-id=18598%3A279512&mode=dev

also updated Answer to "Answer a Question" and "Discussion" to "Other"

Will deploy tomorrow AM

yattias:
> I think the header "Add Bounty" and the other text are OK. I tweaked it slightly, but here are designs for the feature: https://www.figma.com/file/5nYfZkYPjpDQz9cFLeMUVV/Research-Hub?node-id=18598%3A279512&mode=dev
> 
> also updated Answer to "Answer a Question" and "Discussion" to "Other"
> 
> Will deploy tomorrow AM

Sounds good, thank you

PR #1598, opened 2023-10-25
Reference tabs

lightninglu10:
Tab structure for reference manager

Screenshot 2023-10-25 at 2 52 32 PM


components/Document/DocumentViewer.tsx
```diff
@@ -510,6 +513,7 @@ const styles = StyleSheet.create({
   main: {
     position: "relative",
     justifyContent: "center",
+    width: "100%",
```
yattias, line 516: @lightninglu10 Can you make sure this doesn't mess anything up with the document page in regular/full screen mode

PR #1599, opened 2023-10-25
Verified Author

yattias:
Allow users to verify their authorship via LinkedIn or Orcid


components/Home/NavbarRightButtonGroup.tsx
```diff
@@ -74,6 +76,8 @@ function NavbarRightButtonGroup({
     setOpenCaseCounts(totalCount);
   }, [isUserModerator, isUserHubEditor]);
 
+  console.log("user", user.author_profile);
```
lightninglu10, line 79: remove log

PR #1601, opened 2023-10-26
Add Pagination to Hubs

manveerxyz:
Implement pagination on hubs.

## Technical Details
- Created a custom `Pagination` component with custom styling
- Changed `ordering` query param from `score` to `-score` (This is in conjunction with a backend change)
  - `-score` means descending order, which is what we want, we perviously just did some weird stuff on the backend to make this work with `score`.

## Design 
Screenshot 2023-10-26 at 1 50 58 PM

### DIFF
--- a/components/shared/Pagination.tsx
+++ b/components/shared/Pagination.tsx
@@ -1,0 +1,22 @@
+import { styled } from "@mui/material/styles";
+import MuiPagination from "@mui/material/Pagination";
+import colors from "~/config/themes/colors";
+
+const Pagination = styled(MuiPagination)(({ theme }) => ({
+  "& .MuiPaginationItem-outlined": {
+    borderColor: colors.GREY(0.4),
+  },
+  "& .MuiPaginationItem-outlined.Mui-selected": {
+    backgroundColor: colors.NEW_BLUE(0.1),
+    color: colors.NEW_BLUE(),
+    borderColor: colors.NEW_BLUE(0.6),
+  },
+  "& .MuiPaginationItem-outlined.Mui-selected:hover, & .MuiPaginationItem-outlined.Mui-selected.Mui-focusVisible":
+    {
+      backgroundColor: colors.NEW_BLUE(0.2),
+      color: colors.NEW_BLUE(),
+      borderColor: colors.NEW_BLUE(0.6),
+    },
+}));
+
+export default Pagination;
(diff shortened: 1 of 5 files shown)


components/shared/Pagination.tsx
```diff
@@ -0,0 +1,22 @@
+import { styled } from "@mui/material/styles";
```
lightninglu10, line 1: Is it possible to make MUI work with aphrodite? 

Since we have aphrodite everywhere else in the app, it would be nice to keep the same paradigm

yattias: @lightninglu10 I don't think so. It works with emotion or cssjs out of the box. 

@manveerxyz can you run yarn build locally on master and your branch to see if including MUI here impacts the size of the page significantly. Historically we tried to limit MUI to just the ref manager since it has a heavy footprint. If the footprint is not too bad, that's okay

yattias:
@manveerxyz code looks good. I like the look and feel of it too.
How does it look like for mobile?
Also, see my comment above

components/shared/Pagination.tsx
manveerxyz: @yattias `/hubs` is `8KB` on master, and `12.3KB` on my branch.

Master:
```
├ λ /footer                                                1.52 kB         626 kB
├ λ /health                                                422 B           625 kB
├ ● /hubs (ISR: 600 Seconds) (1214 ms)                     8 kB            633 kB
├ ● /hubs/[slug]                                           515 B           683 kB
├ ● /hubs/[slug]/live                                      519 B           683 kB
```

Hub-Pagination:
```
├ λ /footer                                          1.52 kB         626 kB
├ λ /health                                          422 B           625 kB
├ ● /hubs (ISR: 600 Seconds) (6012 ms)               12.3 kB         661 kB
├ ● /hubs/[slug]                                     515 B           683 kB
├ ● /hubs/[slug]/live                                519 B           683 kB
```

manveerxyz: @lightninglu10 I tried using aphrodite but it doesn't work. I used `styled` since we use that in one other place in the codebase to style an MUI component.

Kobe is right that it works with Emotion or classic CSS. The other alternative is to hardcode `styles` in the component, but I'd prefer using the `styled` function from MUI since it seems more standard for just tweaking styling.

manveerxyz:
> @manveerxyz code looks good. I like the look and feel of it too. How does it look like for mobile? Also, see my comment above

Good catch, I centred the pagination bar on mobile so it looks better now.

manveerxyz:
Updated trending hubs query to use `ordering=-paper_count,-discussion_count,id` instead of `ordering=-score`.

lightninglu10:
This looks good, going to merge & deploy it

PR #1603, opened 2023-10-27
Fix bug where loader persisted on profile page

manveerxyz:
**Fix 1:** There is a bug on the profile page where the loader persists on-screen and doesn't go away.

After debugging I found that the `router.events.on("routeChangeError")` would get called with `"Error: Cancel rendering route ..."`, and `routeChangeComplete` would not get called as a result. So the `showLoader.current` timer wouldn't get cleared and would dispatch a `MessageActions.showMessage({ show: true, load: true })` action.

The reason why we get that error is because we make route changes in two places:

1. `router.push` in [handleTabClick](https://github.com/ResearchHub/researchhub-web/blob/master/pages/user/%5BauthorId%5D/%5BtabName%5D/index.js#L228-L252)

2. using the [href tag](https://github.com/ResearchHub/researchhub-web/blob/master/pages/user/%5BauthorId%5D/%5BtabName%5D/index.js#L169) in the HorizontalTabBar

So one of the route changes gets cancelled.

I'm hesitant to change the routing in-case it breaks something, and it also seemed to make sense to hide the loader and clear the timeout if we encounter a `routeChangeError`, and once I made that change the problem was fixed.

I tested other routes and the loader correctly shows/hides in other cases.

---

**Fix 2:** There was a minor bug where the `user/:id/authored-papers` page would say "null" for the date
- I made the `formatTimestamp` func return `""` since this accounts for all its usages

### DIFF
--- a/components/Author/Feed/utils/AuthorFeedUtils.js
+++ b/components/Author/Feed/utils/AuthorFeedUtils.js
@@ -30,7 +30,8 @@ export const getNewestCommentTimestamp = (discussionItem) => {
 
 export const formatTimestamp = (timestamp) => {
   if (!timestamp) {
-    return null;
+    // we return "" so that `null` or `undefined` doesn't show on the UI by accident
+    return "";
   }
   const date = new Date(timestamp);
   return timeAgo.format(date);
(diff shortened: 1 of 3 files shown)


lightninglu10 (approved):

lightninglu10:
Ah good find. LGTM.

I think we should resolve the underlying issue of double pushing too next week.

PR #1604, opened 2023-10-27
Bug Fix: Reset query param when changing sorting on hubs

manveerxyz:
Minor bug where query param wasn't getting reset to page 1 when sorting was changed on hubs.


lightninglu10 (approved):
LGTM

PR #1607, opened 2023-11-03
Fix Reference Metadata

lightninglu10:
Fixes many points in reference metadata

### DIFF
--- a/components/Document/DocumentViewer.tsx
+++ b/components/Document/DocumentViewer.tsx
@@ -534,7 +534,7 @@ const styles = StyleSheet.create({
     width: "100%",
     height: "100%",
     marginTop: 0,
-    overflow: "scroll",
+    overflow: "auto",
   },
   uploadFile: {
     height: "100%",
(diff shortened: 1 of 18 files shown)


components/Document/DocumentViewer.tsx
```diff
@@ -534,10 +534,11 @@ const styles = StyleSheet.create({
     width: "100%",
     height: "100%",
     marginTop: 0,
-    overflow: "scroll",
+    overflow: "auto",
```
yattias, line 537: Can you check commenting works okay in viewer? Also can you check different resolutions just to make sure.

yattias (approved):
Looks good. Just one comment

components/Document/DocumentViewer.tsx
lightninglu10: yes it's fine -- we should always have `overflow: "auto"` over `overflow: "scroll"` -- `overflow: "scroll"` will show the scrollbars always, but auto only shows the scrollbars when you scroll

PR #1611, opened 2023-11-09
Save to reference manager

yattias:
https://www.loom.com/share/70044b5a813b49c996ce94171f63fc21

### DIFF
--- a/components/ReferenceManager/references/reference_organizer/ReferenceProjectsUpsertModal.tsx
+++ b/components/ReferenceManager/references/reference_organizer/ReferenceProjectsUpsertModal.tsx
@@ -45,6 +45,7 @@ export default function ReferenceProjectsUpsertModal({
     resetContext,
     setProjectValue,
     upsertPurpose,
+    redirectAfterUpsert,
   } = useReferenceProjectUpsertContext();
 
   const { resetProjectsFetchTime, setActiveProject, flattenCollaborators } =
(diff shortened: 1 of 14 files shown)


components/Verification/VerificationForm.tsx
```diff
@@ -387,7 +387,7 @@ const VerificationFormSelectProfileStep = ({
   }>({ count: 0, perPage: 0 });
   const [currentPage, setCurrentPage] = useState<{ num: number }>({ num: 1 });
   const [isVerifying, setIsVerifying] = useState(false);
-  const [name, setName] = useState(providerDataResponse?.name || "");
+  const [name, setName] = useState("Jeffrey Koury");
```
lightninglu10, line 390: remove hardcoded jeffrey koury

components/ReferenceManager/references/reference_organizer/context/ReferenceProjectsUpsertContext.tsx
```diff
@@ -55,10 +59,12 @@ export const useReferenceProjectUpsertContext =
 
 export function ReferenceProjectsUpsertContextProvider({ children }) {
   const [isModalOpen, setIsModalOpen] = useState<boolean>(false);
+  const [redirectAfterUpsert, setRedirectAfterUpsert] = useState<boolean>(true);
```
lightninglu10, line 62: Do we make sure we set this back to true if we go into the reference manager?

components/ReferenceManager/references/reference_organizer/ReferenceProjectsUpsertModal.tsx
```diff
@@ -103,6 +105,10 @@ export default function ReferenceProjectsUpsertModal({
         handleCloseModal();
         setIsLoading(false);
 
+        if (!redirectAfterUpsert) {
```
lightninglu10, line 108: Instead of adding this, is it better to add a Prop to this modal?

PR #1614, opened 2023-11-13
Add ability to remove PDF from citation

manveerxyz:
Decided to just add a "Remove" button within the existing "Edit Metadata" menu since this is where users would go to edit the rest of the citation's contents.

It also simplified the UX/implementation.

### No Attachment
no-attachment

### Has Attachment
attachment


yattias (approved):
This looks good

PR #1615, opened 2023-11-14
Allow users to drag and drop .bib files in reference manager

manveerxyz:
This was discussed in last Friday's meeting with Brian.

We wanted to allow users to drag and drop BibTeX files into the main view, to increase discoverability and usability.


yattias (approved):
Looks good

PR #1619, opened 2023-11-20
Fix bug where abstract couldn't be updated in paper page

manveerxyz:
There was a runtime bug where `paper.hubs` was undefined, so trying to call `.map` threw an error.

After looking into it, it doesn't seem like we need to pass `hubs` in the request body (correct me if I'm wrong).

I tried updating abstracts on papers with and without hubs and the request works successfully, and hubs remain unchanged.


lightninglu10 (approved):

lightninglu10:
Looks good -- probably a legacy thing from hubs v1

PR #1620, opened 2023-11-21
Add link to public page in reference manager

manveerxyz:
### Screenshots

Screenshot 2023-11-21 at 9 27 21 AM

Screenshot 2023-11-21 at 9 27 31 AM


config/types/root_types.ts
```diff
@@ -228,6 +228,28 @@ export const parseOrganization = (raw: any): Organization => {
   };
 };
 
+export const generatePageURLForUnifiedDocument = (
```
yattias, line 231: We have a function called `getUrlToUniDoc` in routing.js which does this

yattias (commented):
@manveerxyz was thinking about it and I think we should use a globe icon because we use it to indicate "public page" already when leaving an inline comment

image


I think the current icon implies visiting an external page like the journal web page

config/types/root_types.ts
manveerxyz: Ah gotchya, updated to use that one instead.

lightninglu10:
> @manveerxyz was thinking about it and I think we should use a globe icon because we use it to indicate "public page" already when leaving an inline comment
> 
> image
> 
> I think the current icon implies visiting an external page like the journal web page

Could also use a ResearchHub beaker icon? Or maybe get some new icon designed eventually with the Beaker + open internally

manveerxyz:
> > @manveerxyz was thinking about it and I think we should use a globe icon because we use it to indicate "public page" already when leaving an inline comment
> > image
> > I think the current icon implies visiting an external page like the journal web page
> 
> Could also use a ResearchHub beaker icon? Or maybe get some new icon designed eventually with the Beaker + open internally

That's actually a pretty good idea. The globe can also be confused as a "webpage", but the ResearchHub Beaker says "ResearchHub page" to me.

@yattias thoughts?

PR #1624, opened 2023-11-27
Hide pagination when searching hubs & update remaining Twitter logos

manveerxyz:
Hide pagination when searching hubs.

Update remaining Twitter logos to X. (this was bothering me)
- Kept copy as "X / Twitter"  to limit confusion

### DIFF
--- a/pages/hubs/index.tsx
+++ b/pages/hubs/index.tsx
@@ -256,28 +256,35 @@ const HubsPage: NextPage<Props> = ({
         </ReactPlaceholder>
         {noSuggestionsFound && <div>No hubs found.</div>}
       </div>
-      <div className={css(styles.pagination)}>
-        <Pagination
-          count={Math.ceil(count / 40)}
-          variant="outlined"
-          shape="rounded"
-          color="primary"
-          page={page}
-          // limiting boundary and sibling count reduces width of entire component,
-          // which is useful to prevent overflow/wrappping on mobile.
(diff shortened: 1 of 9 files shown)


pages/hubs/index.tsx
```diff
@@ -256,28 +256,35 @@ const HubsPage: NextPage<Props> = ({
...
-            fetchHubs();
-          }}
-        />
-      </div>
+      {/*
+      Only show the Pagination component if user isn't searching.
+      We only show the first 25 search results (as of now), so pagination is not needed.
+      -> Permalink to 25 limit: https://github.com/ResearchHub/researchhub-backend/blob/e214e56b51ca707ae2d748830e298410f385b299/src/search/views/hub_suggester.py#L41
+      */}
+      {query.length <= 0 && (
```
lightninglu10, line 264: Should we recalculate the number of pages here instead? 

We may want to update the backend to return more than 25 limit

lightninglu10: Then we can show just page 1 if there aren't more pages

manveerxyz: I'm not sure if anybody would need more than 25 items when searching, I'd bet that they'd just modify their search query to be more specific instead.

Adding pagination to searching seems like a bigger lift for not much UX improvement.

yattias (approved):

PR #1634, opened 2023-12-04
[Live Feed] Filters in live feed

yattias:
- Adding filters in live feed mainly so that we can have a place where users can view all bounties and all peer reviews
- Requires filtering mechanism on the BE before this can be merged

### DIFF
--- a/components/Hubs/HubSelect.tsx
+++ b/components/Hubs/HubSelect.tsx
@@ -33,6 +33,7 @@ type HubSelectProps = {
   handleAddHub?: Function;
   handleClick?: (event) => void;
   withPagination?: boolean;
+  maxCardsPerRow?: number;
 };
 
 const HubSelect = ({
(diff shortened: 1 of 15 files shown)


components/Hubs/HubSelect.tsx
yattias, line 1: This is extracted from `pages/hubs/index.tsx` so it can be used in other contexts

manveerxyz:
Tabs look great! Very clean. Overall code looks good as well.

Wondering if we should show the loading placeholders when switching tabs / waiting for API response to give the user a sense of state.

Also I'm honestly not the biggest fan of the Hubs dropdown, it feels a little rough around the edges; e.g. modal has 4-column loading state but 2-column cards, no spacing between search/filter on modal, the "close"/"X" on selected hubs looks off.

We could ship it if it's too much time to fix, but wanted to voice the concern.

PR #1641, opened 2023-12-07
feat: implement notebook feature using lexical

eimenhmdt:


components/Lexical/plugins/EquationsPlugin/index.tsx
yattias, line 1: @eimenhmdt Just for my understanding - I see a bunch of plugins in this directory. Is it because you needed to extend the default plugin functionality? Trying to understand if going forward Lexical plugins will require similar modification or rather only specific ones

components/Lexical/utils/isMobileWidth.ts
yattias, line 1: This probably needs to be removed

components/Lexical/utils/setFloatingElemPosition.ts
yattias, line 1: Curious if this needed to be imported as a file as opposed to being available as part of a library

components/Lexical/ui/Button.tsx
yattias, line 1: Wondering the reasoning behind creating a bunch of custom components vs reusing our own? Can these components you created be used outside of the notebook?

yattias:
@eimenhmdt I see on staging that the CK Editor / Lexical are side by side. I guess this is intentional for you to test things out?

yattias:
@eimenhmdt was there any backend code?

PR #1642, opened 2023-12-08
<HubSelect /> In a couple of legacy spots

yattias:
Fix a couple of spots where the hubs dropdown doesn't show
1. In Editors panel
2. In leaderboard

https://www.loom.com/share/46163186af3f44bfb5722dabf43e781d

### DIFF
--- a/components/Form/FormSelect.js
+++ b/components/Form/FormSelect.js
@@ -10,6 +10,30 @@ import { isDevEnv } from "~/config/utils/env";
 
 const animatedComponents = makeAnimated();
 
+// This is a seldom used control. It should only be used
+// in cases where you do not want click events on your Select Control.
+// e.g. Showing a modal onClick instead of the default dropdown.
+// You will need to define your click event in a wrapper component.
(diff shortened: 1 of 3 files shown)


components/Form/FormSelect.js
```diff
@@ -10,6 +10,30 @@ import { isDevEnv } from "~/config/utils/env";
 
 const animatedComponents = makeAnimated();
 
+// This is a seldom used control. It should only be used
+// in cases where you do not want click events on your Select Control.
+// e.g. Showing a modal onClick instead of the default dropdown.
+// You will need to define your click event in a wrapper component.
+export const CustomSelectControlWithoutClickEvents = ({
```
yattias, line 17: This is not "pretty" but the path of least resistance to override touch and click events on react-select.
It is only used for legacy reasons because some pages have a particular look and feel and are using react-select.

manveerxyz (commented):
I have a couple of nitpicks on the design, but given these aren't a high-visibility locations I'm cool with having this pushed as-is.

PR #1643, opened 2023-12-12
Improper mobile positioning for Hub Select + various modal fixes

yattias:
- Various resolution-specific fixes for post modal
- Bug: Hub Select has incorrect position in mobile
- Updated HubSelect to be more mobile friendly

### Before:
https://www.loom.com/share/cea35ba8291645df81940ab872f26ab2

### After:
https://www.loom.com/share/eb3465116c074c058152fbdab6383c05


manveerxyz (approved):
Definitely looks better than before! I feel like our modals are so broken we probably need a whole refactor in the future.

yattias:
@manveerxyz yeah, the BaseModal needs to be refactored for sure

PR #1646, opened 2023-12-14
Enable Edit/Remove of Questions

yattias:
- Allow users to edit question
- Allow users to remove question
- Removing "Articles" from Live Feed since it introduces terminology complexity
- Created `useCurrentUser` hook

**Menu**
image

**Modal**
image


components/Bounty/BountyWizard.tsx
yattias, line 1: This file is legacy and not used. Will be removed in the future. Keeping it around because it has a few gems we may want to reference.

components/Document/DocumentViewer.tsx
```diff
@@ -89,6 +85,8 @@ const DocumentViewer = ({
   setReferenceItemDatum,
   referenceItemDatum,
   documentViewerClass,
+  withControls = true,
+  allowAnnotations = true,
```
yattias, line 89: Adding new config options because zoom controls and annotations don't make sense on question type

components/Document/api/createOrUpdatePostApi.ts
yattias, line 1: This file was updated and refactored from another function called `createQuestion`. It has some legacy stuff still but more usable than previous version

components/Document/DocumentOptions.tsx
```diff
@@ -49,14 +55,48 @@ function downloadPDF(pdfUrl) {
...
+              // @ts-ignore
+              text: (
+                <div>
+                  {`This ${doc.unifiedDocument.documentType} has been deleted.`}
+                </div>
+              ),
+              buttonText: "OK",
+              onClick: () => {
+                window.location.reload();
+              },
```
manveerxyz, line 92: Shouldn't it redirect somewhere else (e.g. home)? I tried to flow in staging and it didn't seem like the question/post successfully deleted since it just refreshed the page and I still saw the question/post.

I'm also able to still edit the question after it was supposedly "deleted." Perhaps need to clarify what actually happened to the user?

components/Document/api/createOrUpdatePostApi.ts
```diff
@@ -0,0 +1,120 @@
...
+      ? getPlainTextFromMarkdown(payload.textContent)
+      : "",
+    title: payload.title,
+    note_id: payload.noteId,
+    post_id: payload.postId, // Published posts will have a post_id
+  };
+};
+
+// TODO: Kobe: Move this to backend. It isn't the FE responsibility to transmit this event.
+const logEmpEvent = ({
```
manveerxyz, line 45: +1 on the TODO. also small typo I think it should be `logAmpEvent`

components/Document/DocumentOptions.tsx
```diff
@@ -20,13 +25,14 @@ import {
...
-import { StyleSheet, css } from "aphrodite";
+import { StyleSheet } from "aphrodite";
 import colors from "~/config/themes/colors";
 import { useContext } from "react";
 import { DocumentContext } from "./lib/DocumentContext";
 import useCacheControl from "~/config/hooks/useCacheControl";
-import removeDocFromFeatured from "../Admin/api/removeDocFromFeaturedAPI";
 import excludeFromFeed from "../Admin/api/excludeDocFromFeedAPI";
+import censorDocument from "./api/censorDocAPI";
+import { useAlert } from "react-alert";
```
manveerxyz, line 35: Also — don't need to change this now — but I find this alert design kinda ugly and unaligned with our site design.

components/Document/DocumentOptions.tsx
yattias: Agree this should use something like nextjs router. Ill check it out

components/Document/DocumentOptions.tsx
yattias: I don't love it either. We should look into globally updating the style of it. I'll add to our backlog

PR #1648, opened 2023-12-15
Post in ref manager

lightninglu10:
Add ability to put posts in Reference Manager 

- Made all post types available for use for the SaveToRefManager component
- Renders posts in the Reference Manager
- Removed extraneous buttons from Reference Manager for posts
- Depends on https://github.com/ResearchHub/researchhub-backend/pull/1405

Screenshot 2023-12-15 at 11 01 18 AM


manveerxyz (commented):
Nice work! Noticed a couple of bugs/potential-issues.

- Post HTML gets rendered on Metadata sidebar
Screenshot 2023-12-18 at 9 37 12 AM

- I feel like I'd want to see "ResearchHub Post" or even just "Post" as the title instead of "Blog Post"
Screenshot 2023-12-18 at 9 46 34 AM
- On the post tab<br />
1. Upvoting/Downvoting doesn't seem to update the UI<br />
2. Clicking the peer-review stars opens `/post/undefined/undefined/reviews`<br />
3. As a user, I feel like I'd expect some button like "View Full Post" that opens the public page. Because I'm used to seeing comments/reviews/etc. on a post. (not 100% sure)
Screenshot 2023-12-18 at 9 47 53 AM
- Can't create comments because it's calling the `/paper` endpoint
Screenshot 2023-12-18 at 9 51 13 AM

- On the expanded view, the top-right "ResearchHub" button has the wrong link, it points to `/paper/undefined/...`
Screenshot 2023-12-18 at 9 52 01 AM

components/LinkToPublicPage.tsx
yattias, line 1: We currently have the public page CTA in the table view. I think it is sufficient. If you wanted to add the CTA to the tab view then we would need to consider adding it for papers as well for consistency. I have only seen it on posts atm

components/Document/pages/DocumentPageLayout.tsx
```diff
@@ -55,18 +60,24 @@ const DocumentPageLayout = ({
...
       <HeadComponent {...openGraphData.meta} graph={openGraphData.graph}>
         <meta
           name="viewport"
           content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no"
         />
       </HeadComponent>
 
-      <div className={css(styles.topArea)}>
-        <DocumentHeader document={document} metadata={metadata} />
+      <div className={css(styles.topArea) + " " + topAreaClass}>
```
yattias, line 72: I mentioned in slack, I think that the post should have a minimal header component that doesn't have voting, tip, and all the bells and whistles. That way, we can keep the integrity of the `<DocumentHeader>` component

components/Document/pages/DocumentPageLayout.tsx
```diff
@@ -19,12 +18,15 @@ import { useState } from "react";
 
 interface Args {
   document: GenericDocument;
-  errorCode?: number;
-  documentType: DocumentType;
-  tabName?: string;
   children?: any;
   metadata: DocumentMetadata;
   isExpanded?: boolean;
+  noLineItems?: boolean;
```
yattias, line 24: Related to my comment below, I think having a minimal header will avoid us having to configure this component as well

components/Document/DocumentViewer.tsx
```diff
@@ -530,8 +530,8 @@ const styles = StyleSheet.create({
     width: "100%",
   },
   scroll: {
-    overflowX: "scroll",
-    overflowY: "hidden",
+    // overflowX: "scroll",
```
yattias, line 533: Is this safe to remove?

components/Document/DocumentHeaderV2.tsx
```diff
@@ -252,6 +304,14 @@ const styles = StyleSheet.create({
     boxSizing: "border-box",
     justifyContent: "center",
     borderBottom: `1px solid ${config.border}`,
+
+    "@media only screen and (max-width: 767px)": {
```
yattias, line 308: Can we use the predefined breakpoints vars?

components/ReferenceManager/references/ReferencesContainer.tsx
```diff
@@ -1145,6 +1293,17 @@ const styles = StyleSheet.create({
     maxWidth: 400,
     // flex: "1 1 40% !important",
   },
+  headerContentWrapperClass: {
+    paddingLeft: 50,
+
+    "@media only screen and (min-width: 768px)": {
```
yattias, line 1299: Can we use the breakpoint vars here?

yattias (changes requested):

components/Document/pages/DocumentPageLayout.tsx
lightninglu10: I don’t think we should build another component here. The post is a researchhub specific post so it should render very similar to how the post renders on our main site

components/LinkToPublicPage.tsx
lightninglu10: I like having this here, and I think @manveerxyz does as well. On this page it definitely felt like we should have a link to the public page

components/Document/DocumentViewer.tsx
lightninglu10: I believe so, but it changes the page slightly. Having these overflows causes inline comments on the first line to not show up properly

PR #1650, opened 2023-12-20
Bounty + Question 

yattias:
Allow users to add a bounty during question creation process

https://www.loom.com/share/84e75e7f92a24c1a9f0a2d2945731cc0

### DIFF
--- a/components/Bounty/BountyInput.tsx
+++ b/components/Bounty/BountyInput.tsx
@@ -1,0 +1,533 @@
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faUndo } from "@fortawesome/pro-solid-svg-icons";
+import { faClock } from "@fortawesome/pro-regular-svg-icons";
+import { faInfoCircle } from "@fortawesome/pro-light-svg-icons";
+import {
+  BOUNTY_DEFAULT_AMOUNT,
+  BOUNTY_RH_PERCENTAGE,
+  MAX_RSC_REQUIRED,
+  MIN_RSC_REQUIRED,
+} from "./config/constants";
+import { useSelector } from "react-redux";
+import { css, StyleSheet } from "aphrodite";
+import { useState, useEffect } from "react";
+import Bounty from "~/config/types/bounty";
+import colors from "~/config/themes/colors";
+import numeral from "numeral";
+import ReactTooltip from "react-tooltip";
+import ResearchCoinIcon from "../Icons/ResearchCoinIcon";
+import { useExchangeRate } from "../contexts/ExchangeRateContext";
+import { isEmpty } from "~/config/utils/nullchecks";
+import { RootState } from "~/redux";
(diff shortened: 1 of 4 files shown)


components/Bounty/BountyInput.tsx
yattias, line 1: This is a refactor I did to extract the "Input part" out of the `BountyModal` component. I have yet to refactor to BountyModal component due to time constraints but that would be the next step here. This is 98% existing code from `BountyModal` component.

components/Question/AskQuestionForm.tsx
```diff
@@ -177,18 +232,65 @@ function AskQuestionForm({ post, user, onExit }: AskQuestionFormProps) {
...
+              <div onClick={() => setWithBounty(false)}>
+                <Button size="small" customButtonStyle={styles.removeBountyBtn}>
+                  <FontAwesomeIcon icon={faPlus} style={{ marginRight: 4 }} />
+                  Remove Bounty
+                </Button>
+              </div>
+            ) : (
+              <div onClick={() => setWithBounty(true)}>
+                <Button size="small" customButtonStyle={styles.addBountyBtn}>
+                  <FontAwesomeIcon icon={faPlus} style={{ marginRight: 4 }} />
```
manveerxyz, line 254: nit: `icon` should be `faTimes` if `withBounty === true`

components/Question/AskQuestionForm.tsx
```diff
@@ -73,6 +77,14 @@ function AskQuestionForm({ post, user, onExit }: AskQuestionFormProps) {
   const [isSubmitting, setIsSubmitting] = useState<boolean>(false);
   const currentUser = useCurrentUser();
   const documentContext = useContext(DocumentContext);
+  const [withBounty, setWithBounty] = useState<boolean>(false);
+  const [bountyOffered, setBountyOffered] = useState<number>(0);
+  const [bountyError, setBountyError] = useState<any>(null);
+  const alert = useAlert();
+  const handleBountyInputChange = ({ hasError, errorMsg, value }) => {
+    setBountyOffered(value);
+    setBountyError(errorMsg);
```
manveerxyz, line 86: it doesn't seem like we show the error message anywhere, e.g. if I don't have enough balance, or am below the minimum required.

Currently if there's an error in the `BountyInput` I just see a disabled button and have to guess why it's not letting me submit.

manveerxyz (commented):
Looks good for the most part! Just small comments.

components/Question/AskQuestionForm.tsx
yattias: Good call

components/Question/AskQuestionForm.tsx
yattias: Thanks for letting me know. I did a rebase and I think that might have gotten messed up. There should be an error in there.

PR #1654, opened 2024-01-03
Setup Preregistration Crowdfunding

manveerxyz:
Implement UI/UX for preregistration fundraising.

Loom Demo: https://www.loom.com/share/6815d644055b4026a7684c62730247cd


yattias:
Excellent job on this feature. Product wise, I struggled to find anything wrong with it while doing QA.

Here are some minor observations:

- Can we display a clearer error message when user tries to support their own fundraise? Right now it is “something went wrong”
- Saving to ref manager works but throws an error when you try to view fundraise in ref manager
- Let’s add apply for funding since after we launch this, the graphic on the right will go away. Users will still need to be able to apply.  I think we can mimic what is on hub pages:
Pasted Graphic 3

- This is not you but I think something got messed up with the tabs. If it’s quick, would be great if you could take a look. There is some padding added - My guess is that it is when Pat added functionality to view posts in ref manager.

Pasted Graphic 2

- This can wait but should render the fundraising component in profile page. I think we can do it the following sprint.
Pasted Graphic 1
- Can we Indent for small screens (15px on each side)? Right now the component stretches to 100% but I think for consistency would be great to be able to draw an imaginary line from the left of the title all the way down the page and have only white space. I notice you do it in the list view.
Pasted Graphic

Thanks!

components/Fundraise/FundraiseCard.tsx
```diff
@@ -0,0 +1,273 @@
...
+  onUpdateFundraise?: (fundraise: Fundraise) => void;
+};
+
+const FundraiseCard = ({
+  fundraise: f,
+  showSupportButton = true,
+  onUpdateFundraise = () => {},
+}: FundraiseCardProps): ReactElement => {
+  const barFillAmount = useMemo(() => {
+    const fillAmount = (f.amount_raised.rsc / f.goal_amount.rsc) * 100;
```
yattias, line 33: If you'd like to one-line it you can: `Math.min((f.amount_raised.rsc / f.goal_amount.rsc) * 100, 100)`.
Fine as is though.

components/Fundraise/FundraiseCard.tsx
```diff
@@ -0,0 +1,273 @@
...
+            width={18}
+            height={18}
+          />
+          <div
+            className={css(
+              styles.amountRaisedText,
+              isFundraiseFulfilled(f) && styles.amountRaisedTextFulfilled
+            )}
+          >
+            {numeral(f.amount_raised.rsc).format("0[,]0")} RSC
```
yattias, line 61: We have a function called `formatBountyAmount`. Wonder if we can repurpose it and use it here for consistency.

components/Fundraise/FundraiseCard.tsx
```diff
@@ -0,0 +1,273 @@
...
+
+export type FundraiseCardProps = {
+  fundraise: Fundraise;
+
+  showSupportButton?: boolean;
+  onUpdateFundraise?: (fundraise: Fundraise) => void;
+};
+
+const FundraiseCard = ({
+  fundraise: f,
```
yattias, line 28: I realize that `fundraise` is a prop on unified_document and that the parent component `<FeedCard>` is very TS unfriendly. However, for the sake of creating rock solid components on a go-forward basis, we can create a `Fundraise` type + a parser for it. We can then invoke it in `FeedCard` only for that particular property so that `FundraiseCard` and its data are TS friendly. In `DocumentHeader`, you can update the `DocumentMetadata` type to include a `fundraise: Fundraise` prop

components/Fundraise/api/contributions.ts
```diff
@@ -0,0 +1,76 @@
+import { Helpers } from "@quantfive/js-web-config";
+import API, { generateApiUrl } from "~/config/api";
+import { ID } from "~/config/types/root_types";
+import { captureEvent } from "~/config/utils/events";
+import { isNullOrUndefined } from "~/config/utils/nullchecks";
+import { Purchase, parsePurchase } from "~/config/types/purchase";
+import { Fundraise, parseFundraise } from "../lib/types";
+
+export const fetchFundraiseContributions = async ({
```
yattias, line 9: To make it super easy to understand whether a function is utility vs api call, we suffix API functions with `Api` suffix. It reads a little funny but it gets the job done. Could you update the API functions here:
e.g. `fetchFundraiseContributionsApi`

components/Fundraise/FundraiseCard.tsx
yattias: On second look , it appears you already create a type below but the snake case threw me off. Can you update to camel case for consistency? The nice thing about camelcase is that it is a visual queue that an object literal has been converted to TS and been "dealt with".

config/hooks/useRefreshUserBalance.ts
yattias, line 1: 💯  Nice job creating this hook. Thanks.

yattias (commented):
Looks great. Left very minor comments

components/Fundraise/FundraiseCard.tsx
manveerxyz: Ah good point, that's a decent convention to have.

yattias (approved):

PR #1655, opened 2024-01-04
Displaying user nav in mobile

yattias:
# Before
image

# New: Standard user

image

# New: Admin user

image

Inspired by: Notion, Medium, Github

image

### DIFF
--- a/components/Home/NavbarRightButtonGroup.tsx
+++ b/components/Home/NavbarRightButtonGroup.tsx
@@ -109,6 +109,7 @@ function NavbarRightButtonGroup({
           className={css(
             styles.buttonsGroup,
             styles.borderNone,
+            styles.avatarContainerGroup,
             styles.noPaddingLeft
           )}
         >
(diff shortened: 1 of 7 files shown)


manveerxyz:
I tried it on my phone and had a couple of comments.

1. It feels a little cluttered to me. Specifically:
    1. The profile icon has more right-padding than the hamburger menu's left-padding
    2. The spacing between the icons doesn't seem consistent, and the vertical line honestly looks a little off — not sure if it'd look better with/without though
1. The notification modal overflows

yattias:
> Good idea to add the rest of the top-bar features along w the profile icon! I tried it on my phone and had a couple of comments.
> 
> 1. It feels a little cluttered to me. Specifically:
>    
>    1. The profile icon has more right-padding than the hamburger menu's left-padding
>    2. The spacing between the icons doesn't seem consistent, and the vertical line honestly looks a little off — not sure if it'd look better with/without though
> 2. The notification modal overflows
> 
> 

- Good catch on the overflow. I'll take a look.
- I'll double check the spacing between icons. I measured using a ruler and was consistent.
- I think the messiness comes from having one icon too many. As it stands, I do think we need to keep the number of icons as is because notification and RSC have dropdowns as opposed to redirect to different pages
- I did try removing the vertical bar and it looked a bit messier, especially with more icons:

image

PR #1660, opened 2024-01-08
Notification related Changes; Toast;

yattias:
- Refactored toast notification so that it can be used from anywhere in the app. Prior to this, it was only scoped to paper upload flow
- Shortened Notification in dropdown. Truncate if beyond 100 characters
-  Added toast to save to reference manager flow.
- Optimistic Save in save to ref manager dialog

https://www.loom.com/share/6cf1fc5569064d72b9226b7a303255e1

### DIFF
--- a/components/Notifications/lib/showSaveToRefManagerToast.tsx
+++ b/components/Notifications/lib/showSaveToRefManagerToast.tsx
@@ -1,0 +1,117 @@
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import Link from "next/link";
+import IconButton from "~/components/Icons/IconButton";
+import { StyleSheet, css } from "aphrodite";
+import showToast from "./showToast";
+import { faCheckCircle } from "@fortawesome/pro-solid-svg-icons";
+import { Organization } from "~/config/types/root_types";
+import colors from "~/config/themes/colors";
+
+interface Props {
+  action: "ADD" | "REMOVE";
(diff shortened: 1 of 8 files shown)


components/Notifications/lib/showSaveToRefManagerToast.tsx
yattias, line 1: Specific toast for saving to ref manager. In future, we may be able to refactor and reuse

manveerxyz: I've started using chakra-ui at Scholar and love the interfaces/composability of their components. Perhaps can draw some inspiration when we do decide to create a generalized component:
- Toast: https://chakra-ui.com/docs/components/toast/usage#usage
- Modal: https://chakra-ui.com/docs/components/modal#usage

manveerxyz (commented):
Overall looks good. I don't have many strong opinions on the toast design after looking at it again, just a comment on the transition below.

yattias:
@manveerxyz agree the animation feels odd. I updated to a fade-in/out effect which looks much nicer

manveerxyz:
> @manveerxyz agree the animation feels odd. I updated to a fade-in/out effect which looks much nicer

I like it! Only thing is now it doesn't seem like the toast goes away unless I click it.

yattias:
Good catch. @manveerxyz just pushed a fix for it.

PR #1666, opened 2024-01-12
Remove hub selection and ORCiD connect from onboarding

manveerxyz:
- Remove hub selection and ORCiD connect modal
- Fix bug with ORCiD connect in profile page
- Switch to getting/setting new `has_completed_onboarding` field since `has_seen_orcid_connect_modal` field didn't make sense anymore.
- Update "Complete your profile" page
- Added support for Google Scholar links and removed Facebook links
- Update modal that user gets shown on first vote/contribution
- Tested on Mobile and Desktop

### Screenshots
Screenshot 2024-01-15 at 4 59 08 PM

Screenshot 2024-01-15 at 5 25 49 PM


yattias:
Profile form looks great! A few relatively small things I noticed

- Maybe bug: Specifying links in the onboarding flow didn't work for me. I tried putting in links like meow.com

- Maybe bug: Orcid click prompts a connect mechanism. I think it should work like the other links - maybe connect is overkill here. Also, clicking on another user's orcid button brings up orcid connect

- Padding in mobile: There is a left padding that appears in mobile. Could we add a corresponding right padding to even it out?
image

- Can we add a pencil or have the "update" hover state always appear? Reason is that it is not clear you can update the picture unless you interact with it
image

- Tiny: Instead of twitter.com placeholder, can we have x.com?
image

components/Onboard/OnboardForm.js
```diff
@@ -422,6 +433,39 @@ class OnboardForm extends Component {
...
+              placeholder="https://linkedin.com/in/username"
+            />
+            <FormInput
+              id="twitter"
+              onChange={this.handleFormChange}
+              containerStyle={[styles.formInput, styles.socialInput]}
+              value={this.state.twitter}
+              icon={<FontAwesomeIcon icon={faSquareXTwitter} />}
+              iconStyles={[styles.socialIcon, styles.xTwitterIcon]}
+              placeholder="https://twitter.com/username"
```
yattias, line 457: Maybe x.com instead of twitter as placeholder?

components/Onboard/api/setCompletedOnboardingApi.ts
```diff
@@ -0,0 +1,19 @@
+import { Helpers } from "@quantfive/js-web-config";
+import API, { generateApiUrl } from "~/config/api";
+import { captureEvent } from "~/config/utils/events";
+
+export const setCompletedOnboardingApi = async (): Promise<void> => {
```
yattias, line 5: Would the user see onboarding form again if they just don't click continue or skip but rather just visit the homepage? Ideally they should not

manveerxyz: Good point, updated it to auto-mark onboarding as complete if the user navigates away.

components/Onboard/OnboardForm.js
manveerxyz: updated!

PR #1658, opened 2024-01-08
Improve how we handle copyrighted PDFs

manveerxyz:
### Updates
- Update copy when PDF isn't viewable
- Use new `pdf_copyright_allows_display` flag from backend to determine whether to enable PDF Upload
- User-uploaded PDF shows in reference manager, even if paper has copyright restrictions
- Saving a paper to the reference manager with a copyrighted PDF won't copy the PDF to the citation
- Let moderators remove PDF from moderator panel
- Let moderators edit license metadata

### Demo
https://www.loom.com/share/e5681b0ccff342f783c5a91c1bbf8835?sid=c0dcf2e2-29f5-4c61-8d26-36fc00e0b9b3

### Test Papers
- Should Show PDF: `oa_status = null` and `pdf_license = cc-by`: https://researchhub-web-git-manveer-pdf-licenses-researchhub.vercel.app/paper/131491/the-lbca-lipoprotein-and-ctpa-protease-assemble-an-oligomeric-complex-to-control-peptidoglycan-hydrolases-in-ipseudomonas-aeruginosai
- Should _NOT_ Show PDF: `oa_status = null` and `pdf_license = null`: https://researchhub-web-git-manveer-pdf-licenses-researchhub.vercel.app/paper/131736/pamlr-a-toolbox-for-analysing-animal-behaviour-using-pressure-acceleration-temperature-magnetic-and-light-data-in-r
- Should Show PDF: `oa_status = gold` and `pdf_license = null`: https://researchhub-web-git-manveer-pdf-licenses-researchhub.vercel.app/paper/131836/formation-of-a-traditional-chinese-medicine-self-assembly-nanostrategy-and-its-application-in-cancer-a-promising-treatment

PR #1671, opened 2024-01-19
Fix bug where null href caused paper page to crash

manveerxyz:
There were a bunch of papers failing to render on the server side because of a `null` value in a `href` tag.

It seems like this may have been caused by `paper.url = null`, and may have newly been surfaced as an issue when we made the change to copyright handling (https://github.com/ResearchHub/researchhub-web/pull/1658).

### DIFF
--- a/components/Document/DocumentLineItems.tsx
+++ b/components/Document/DocumentLineItems.tsx
@@ -118,7 +118,9 @@ const DocumentLineItems = ({
                 href={
                   isPaper(document) && document.externalUrl
                     ? document.externalUrl
-                    : `https://` + document.doi
+                    : document.doi.startsWith("http")
+                    ? document.doi
+                    : `https://doi.org/${document.doi}`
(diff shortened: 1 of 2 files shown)


components/Document/DocumentLineItems.tsx
```diff
@@ -118,7 +118,9 @@ const DocumentLineItems = ({
                 href={
                   isPaper(document) && document.externalUrl
                     ? document.externalUrl
-                    : `https://` + document.doi
+                    : document.doi.startsWith("http")
```
manveerxyz, line 121: There was also an issue where the DOI field on the paper page added an extra `https://`; e.g. `https://https://doi.org/10.1364/OFC.2022.W1H.1`

yattias: I see. look okay

yattias (approved):

PR #1674, opened 2024-01-23
Update nvm instructions in readme to use nvmrc file config

tomerds:
as per title


jo-elimu (changes requested):
Is it necessary to use `nvm` at all? In my case, the setup works with `yarn install` alone.

PR #1675, opened 2024-01-26
Fix bug where all papers were marked as closed access or non-commercial

manveerxyz:
Tested locally with data from papers that were flagged by the team and realized there was a logic issue.

### DIFF
--- a/components/Paper/Tabs/PaperTab.js
+++ b/components/Paper/Tabs/PaperTab.js
@@ -154,7 +154,7 @@ function PaperTab(props) {
       );
     }
     if (
-      !paper.pdf_copyright_allows_display ||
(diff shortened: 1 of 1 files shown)


components/Paper/Tabs/PaperTab.js
```diff
@@ -154,7 +154,7 @@ function PaperTab(props) {
       );
     }
     if (
-      !paper.pdf_copyright_allows_display ||
+      !paper.pdf_copyright_allows_display &&
       !paper.pdfCopyrightAllowsDisplay
```
manveerxyz, line 158: I still want to check both fields for redundancy/robustness purposes, since this is pretty important logic.

yattias (approved):

PR #1676, opened 2024-02-02
Temporary shim to fetch recommendations

yattias:
Added temporary shim to fetch recommendations 

e.g. http://localhost:3000/?experiment=recs&user_id=1

### DIFF
--- a/config/api.js
+++ b/config/api.js
@@ -795,13 +795,30 @@ const routes = (BASE_URL) => {
         frontendFilters: selectedFilters,
         hubId,
       });
+      let url =
+        BASE_URL + "researchhub_unified_document/get_unified_documents/";
+
+      // Kobe, 2024-02-01: This is a temporary shim to allow for recs experiment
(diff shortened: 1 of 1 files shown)


config/api.js
```diff
@@ -795,13 +795,30 @@ const routes = (BASE_URL) => {
...
+          window.location.search.includes("experiment")
+        ) {
+          const userId = window.location.search.split("user_id=")[1];
+          url =
+            BASE_URL +
+            `researchhub_unified_document/recommendations/?user_id=${userId}`;
+        }
+      } catch (error) {
+        console.error("Failed to generate recs url", error);
+      }
```
yattias, line 816: The above is not the nextjs approach of fetching query params and generally contains a bunch of anti-patterns but, it is nevertheless the simplest way of fetching recommendations and displaying them in feed given the overly complex feed hierarchy. I've identified this place as the only place that requires a singular code change to achieve the goal.

PR #1681, opened 2024-02-13
use node v18 instead of v14

lightninglu10:
Node v18 instead of v14 in the readme


gzurowski:
This PR is superseded by #1694

PR #1684, opened 2024-02-16
Implement V1 of For You Feed

manveerxyz:


components/Author/Feed/types/forYouFeedTypes.ts
```diff
@@ -0,0 +1,91 @@
...
+  hubs: Hub[];
+  unifiedDocument: UnifiedDocument;
+
+  userVote: number | null;
+  score: number;
+
+  createdBy: RHUser;
+  createdDate: string;
+}
+export interface ForYouFeedPaperItem extends ForYouFeedItem {
```
yattias, line 30: Maybe we should use more generic terms like `FeedPaperItem`. Reason being is that this JSON structure will be returned in other areas of the site as well. For example, I intend to update the live feed, profile feed, ref manager feed, ... to include data that conform to this structure.

components/Author/Feed/types/forYouFeedTypes.ts
```diff
@@ -0,0 +1,91 @@
...
+  item: Paper;
+}
+export interface ForYouFeedPostItem extends ForYouFeedItem {
+  item: Post;
+}
+export interface ForYouFeedBountyItem extends ForYouFeedItem {
+  item: Bounty;
+}
+export interface ForYouFeedCommentItem extends ForYouFeedItem {
+  item: ForYouFeedCommentObject;
```
yattias, line 40: My guess is that you created this type because some data is missing? If so, I can include the additional missing so that we don't have to create this new type

yattias (approved):

components/Author/Feed/types/forYouFeedTypes.ts
manveerxyz: Yeah that is why, it didn't fit within the Comment parser or the CommentContribution parser

pages/recs/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const authToken = cookies[AUTH_TOKEN];
+  const defaultProps = getFetchDefaults({ query, authToken });
+
+  if (process.browser) {
+    return defaultProps;
+  }
+
+  try {
+    const selectedFilters = getSelectedUrlFilters({ query, pathname: "/" });
+    const initialFeed = await fetchUnifiedDocFeed(
```
yattias, line 35: Can we remove this for now? I think it is causing the serverless function to crash

PR #1692, opened 2024-03-07
Sanitize post html before rendering

koutst:
Sanitize post html before rendering to avoid XSS attacks. I decided to just use the default attributes listed here https://github.com/apostrophecms/sanitize-html?tab=readme-ov-file#default-options plus video and tested it by creating a note with each of the rich editing features. 

We are not sanitizing the notebooks on creation because we will be migrating away from CKEditor so we will address those changes along with a sanitization of any existing notes.

### DIFF
--- a/components/Document/api/fetchPostFromS3.ts
+++ b/components/Document/api/fetchPostFromS3.ts
@@ -15,15 +15,27 @@ const fetchPostFromS3 = async ({ s3Url, cleanIntroEmptyContent = true }: Props):
         msg: `Error fetching post from S3`,
       });
     }
-    let _html = await response.text();
-    _html = sanitizeHtml(_html, {
-      allowedTags: sanitizeHtml.defaults.allowedTags.concat(['video', 'source']),
-      allowedAttributes: {
-        ...sanitizeHtml.defaults.allowedAttributes,
-        video: ['src', 'type', 'controls', 'autoplay', 'muted', 'loop', 'width', 'height'],
-        source: ['src', 'type']
-      }
-    });
+    let _html = await response.text().then((text) =>
+      sanitizeHtml(text, {
+        allowedTags: sanitizeHtml.defaults.allowedTags.concat([
+          "source",
+          "video",
+        ]),
+        allowedAttributes: {
+          ...sanitizeHtml.defaults.allowedAttributes,
+          video: [
+            "autoplay",
+            "controls",
+            "height",
+            "loop",
+            "muted",
+            "src",
+            "type",
+            "width",
+          ],
+          source: ["src", "type"],
+        },
+      }));
 
     if (cleanIntroEmptyContent) {
       // CK Editor saves H1 tags on S3.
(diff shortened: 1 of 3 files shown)


components/Document/api/fetchPostFromS3.ts
```diff
@@ -15,6 +16,14 @@ const fetchPostFromS3 = async ({ s3Url, cleanIntroEmptyContent = true }: Props):
       });
     }
     let _html = await response.text();
+    _html = sanitizeHtml(_html, {
```
gzurowski, line 19: Nit: Could use `then()` here and format and order list of attributes.

```
    let _html = await response.text().then((text) =>
      sanitizeHtml(text, {
        allowedTags: sanitizeHtml.defaults.allowedTags.concat([
          "source",
          "video",
        ]),
        allowedAttributes: {
          ...sanitizeHtml.defaults.allowedAttributes,
          video: [
            "autoplay",
            "controls",
            "height",
            "loop",
            "muted",
            "src",
            "type",
            "width",
          ],
          source: ["src", "type"],
        },
      })
    );
```

gzurowski (approved):
LGTM 👍

components/Document/api/fetchPostFromS3.ts
```diff
@@ -14,7 +15,29 @@ const fetchPostFromS3 = async ({ s3Url, cleanIntroEmptyContent = true }: Props):
         msg: `Error fetching post from S3`,
       });
     }
-    let _html = await response.text();
+    let _html = await response.text().then((text) =>
```
yattias, line 18: This could work a little different than expected but the main motivation to use `await` is to circumvent `then` and avoid the javascript [Pyramid of Doom](https://blog.stackademic.com/callback-hell-callback-pyramids-in-javscript-d66e6ba27736).

In other words, you may be able to just do: `sanitizeHtml(text, {...` in subsequent line without being scoped to `then`

components/Document/api/fetchPostFromS3.ts
```diff
@@ -14,7 +15,29 @@ const fetchPostFromS3 = async ({ s3Url, cleanIntroEmptyContent = true }: Props):
...
+      sanitizeHtml(text, {
```
yattias, line 19: Not sure if we need to sanitize these, but in theory, more than just the following three tags can exist as part of the response. For example, `table`, `figure`, ...

koutst: Those are included in the default allowed tags which I'm concatenating with some missing tags https://www.npmjs.com/package/sanitize-html

yattias (approved):
This looks good but before merging, check out the comments I left. Once you take a look, feel free to merge

components/Document/api/fetchPostFromS3.ts
koutst: Maybe I should be more explicit and just define all of them here?

yattias: Ah gotcha. I think if you add a comment, it should be more than sufficient

components/Document/api/fetchPostFromS3.ts
koutst: Ended up reverting to avoid deep nesting as suggested by Kobe.

PR #1698, opened 2024-03-22
[Search] Highlight author names in search suggestions; related refactor;

yattias:
Highlighting author names in search suggestions metadata row

image

### DIFF
--- a/components/SearchSuggestion/lib/util.ts
+++ b/components/SearchSuggestion/lib/util.ts
@@ -1,3 +1,6 @@
+import sanitizeHtml from "sanitize-html";
+import { SanitizedAndSafeHtml } from "~/config/types/root_types";
+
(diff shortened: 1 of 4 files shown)


components/SearchSuggestion/lib/util.ts
```diff
@@ -26,3 +29,27 @@ export const buildPageUrlFromSuggestion = (suggestion: Suggestion) => {
       return `/question/${questionSuggestion.id}/${questionSuggestion.slug}`;
   }
 };
+
```
yattias, line 32: Note the function below returns "Safe" HTML. HTML and not string because we are using javascript replace function to replace portions with highlighted equivalent (this requires HTML)

koutst (approved):
LGTM

PR #1706, opened 2024-03-26
Remove delete user request and admin functionality

koutst:
- The user delete request and admin dashboard are not working properly. We will rework this feature in the future if required.


yattias (approved):
LGTM

PR #1707, opened 2024-03-26
[Search] Various UX fixes

yattias:
- Do not debounce when user presses "Enter" key
- Do not auto-open suggestion drawer on page load
- Close suggestion drawer when notification btn is clicked on
- Remove `external_source` property which applies filter incorrectly and causes results not to show in search results page


koutst (approved):
LGTM

PR #1710, opened 2024-03-28
[Search] Updating design for best results, fix broken filters

yattias:
- Updating visual design of "Best Results" page to match more closely to design
- Fix broken filters: journal, hubs
- Updated layout to work in mobile

More design work is needed and will follow up next sprint.


koutst (approved):
LGTM

PR #1714, opened 2024-04-08
[Search] Adding missing fields to paper card that are now available

yattias:
Fields are now available to include on paper card https://github.com/ResearchHub/researchhub-backend/pull/1521/files


koutst (approved):
LGTM

PR #1715, opened 2024-04-10
[Feed] Save to ref manager on feed card 

yattias:
- Save to ref manager directly in feed card
- Rendering citations, peer review score and others on feed card

Screenshot 2024-04-10 at 2 16 37 PM


components/Author/Tabs/FeedCard.tsx
yattias, line 1: Just as an FYI, this component might be the worst frontend component we have in our codebase and needs to be fully rewritten. Until we do so, the strategy is to add limited code to it and instead, add child components

components/Author/Tabs/FeedCard.tsx
```diff
@@ -285,6 +273,17 @@ function FeedCard({
   const bountyAmount = documentFilter?.bounty_total_amount;
   const hasActiveBounty = documentFilter?.bounty_open;
 
+  let numOfVisibleHubs = 3;
```
yattias, line 276: Ideally this display logic would be done via CSS but given how the component is structured, it is not possible to do so without rewriting or relying on JS.

components/Author/Tabs/FeedCard.tsx
```diff
@@ -434,59 +465,64 @@ function FeedCard({
...
-                      <div className={css(styles.desktop)}>
+
+                      <div className={css(styles.desktop, styles.badges)}>
                         <DocumentHubs
                           hubs={parsedHubs}
                           withShowMore={false}
                           hideOnSmallerResolution={true}
+                          numOfVisibleHubs={numOfVisibleHubs}
                         />
+                        {hasActiveBounty && (
```
yattias, line 476: This is not new code, moved around

components/Feed/FeedCardActivity.tsx
yattias, line 1: This new component renders the feed card activity state (comments, citations, etc..)

components/Search/SearchResultsForDocs.js
```diff
@@ -582,8 +582,9 @@ const SearchResultsForDocs = ({ apiResponse, entityType, context }) => {
             return (
               <FeedCard
                 {...post}
+                unifiedDocumentId={post.unified_document_id}
                 formattedDocType={"post"}
-                key={post?.id || index}
```
koutst, line 586: Is this index fallback not necessary anymore?

koutst (approved):
LGTM

components/Search/SearchResultsForDocs.js
yattias: Yeah, it wasn't necessary to begin with

PR #1716, opened 2024-04-17
[Search] Percentile, license and mobile friendly search filters

yattias:
- Adding license as a filter
- Adding percentile as a filter
- Using `<SwipableDrawer />` and `<Modal>` components to display search filters in smaller resolutions
- While at it, converted `SearchResultsForDocs` component into typescript and extracted filtering logic into `<SearchFiltersForDocs />`  and `SearchFiltersContext`
- Fix back button issue redirecting to wrong place
image

image

image

image


components/Search/SearchFiltersForDocs.tsx
yattias, line 1: This is a new component that was extracted from mega component of `SearchResultsForDocs`

components/Search/lib/AppliedFilters.tsx
yattias, line 1: New component that was extracted from `SearchResultsForDocs` it holds rendering logic for applied filters

components/Search/lib/SearchFiltersContext.tsx
yattias, line 1: The purpose of this context is to hold filtering state that can be used across multiple components

components/Form/SimpleSlider.tsx
```diff
@@ -0,0 +1,43 @@
...
+    setValue(newValue as number);
+    onChange(newValue as number);
+  };
+
+  const handleSliderChange = (event: Event, newValue: number | number[]) => {
+    setValue(newValue as number);
+  };
+
+  function valueLabelFormat(value: number) {
+    return value + "th percentile";
```
koutst, line 25: Nit: This won't work for every percentile. 1st, 2nd, 3rd, 21st, 23rd, 24th etc.

koutst (approved):
LGTM

components/Form/SimpleSlider.tsx
yattias: good call

PR #1717, opened 2024-04-18
[PDF Viewer] Dimension related fixes

yattias:
- Fix: PDF extends beyond viewport in mobile
- Fix: Align PDF controls against the PDF


koutst (approved):
LGTM!

PR #1719, opened 2024-04-22
[Bug] Prevent double click on tips; pass uuid4 transaction_id

yattias:


config/fetch.js
```diff
@@ -326,6 +327,7 @@ export const supportContent = async ({ contentType, objectId, amount }) => {
     content_type: contentType,
     object_id: objectId,
     amount,
+    transaction_id: uuidv4(),
```
koutst, line 330: Would this generate a unique id on each submit? Shouldn't we generate it once on modal load?

koutst (approved):
LGTM

config/fetch.js
yattias: You know, you are indeed right. On modal load is safer. Curious - do you have any plans when it comes to what to do with this field in the backend?

koutst: Just store it and query the db to check if the same client ID has already been used and fail if it has.

PR #1721, opened 2024-04-24
[Save to Ref Manager] Auto pick organization last used

yattias:
- Default to last organization, project user saved reference to


components/contexts/ReferenceManagerSettings.tsx
yattias, line 1: Contexts in react are used as a safe "global storage" of data which can be used in various areas of an app.

gzurowski (approved):

koutst (approved):
LGTM

PR #1724, opened 2024-04-29
[Search] Couple of bug fixes

yattias:
- Suggestions drawer not closing after pressing "Enter"
- Highlights in paper results cause certain fields like authors to not render

### DIFF
--- a/config/utils/string.js
+++ b/config/utils/string.js
@@ -28,7 +28,8 @@ export const isString = (str) => {
 };
 
 export const stripHTML = (str) => {
-  return (str || "").replace(/(<([^>]+)>)/gi, "");
+  if (typeof str !== "string") return str;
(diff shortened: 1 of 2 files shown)


config/utils/string.js
```diff
@@ -28,7 +28,8 @@ export const isString = (str) => {
 };
 
 export const stripHTML = (str) => {
-  return (str || "").replace(/(<([^>]+)>)/gi, "");
+  if (typeof str !== "string") return str;
```
koutst, line 31: What is being passed in instead of a string?

yattias: In some cases, such as when we highlight keywords in search, react objects could be passed in.
I don't like this approach I did here. I'll try to move this logic a bit upward which may be a bit verbose but better suited

PR #1750, opened 2024-06-13
fix: hub dropdown empty label

stackedq:
fixes https://github.com/ResearchHub/issues/issues/1


yattias:
@stackedq still getting empty hubs when editing a post

image

yattias:
After testing a couple of times, I ended up pushing a fix. Your code was close but a bit off.
Merged your z-index issue PR though.

PR #1752, opened 2024-06-13
Point staging to new prod

gzurowski:
This is for testing the new AWS prod environment where we point staging to the new prod backend API.


yattias (approved):
Blessed

PR #1755, opened 2024-06-13
Post quick fix

lightninglu10:
Quick fix for: https://github.com/ResearchHub/issues/issues/4

This isn't 100% the right way to fix the issue. The reason the issue is happening is because we're rendering raw HTML, which doesn't work 100% with CKEditor.

We should be rendering a readonly instance of CKEditor inside the `<DocumentViewer />`. This will take longer to get right because there are quirky things about CKEditor. The reason we should be rendering CKEditor inside the DocumentViewer is because CKEditor does some kind of postprocessing on the HTML.

### DIFF
--- a/components/CKEditor/ELNEditor.js
+++ b/components/CKEditor/ELNEditor.js
@@ -138,6 +138,8 @@ const ELNEditor = ({
     };
   };
 
+  console.log(currentNote.latest_version?.src);
+
   return (
     <div className={css(styles.container)}>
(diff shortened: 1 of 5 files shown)


yattias (approved):

yattias:
@lightninglu10 I merged this but removed the dynamic loading of components for now because I don't have testing capacity to ensure that doesn't hurt the UX.

PR #1763, opened 2024-06-26
Avoiding crashing the whole moderator dash if error is thrown on claim

yattias:
- Wrap each claim processing with try/catch to avoid crashing the mod dash


koutst (approved):
LGTM

PR #1764, opened 2024-06-28
Fix issue with adding valid DOIs to ref manager

kerkelae:
This PR aims to solve https://github.com/ResearchHub/issues/issues/34.

The issue seems to have been `issn.join` failing if `issn` is `null` or not an array. The fix involves adding a check to ensure `issn` is an array before calling the `join` method. The updated code uses `Array.isArray(issn) ? issn.join(", ") : ""`, which checks if `issn` is an array; if it is, it joins the array elements into a string separated by commas. If `issn` is not an array or is `null`, it defaults to an empty string. This prevents the application from throwing an error when trying to call `join` on a `null` value.

I ran the code locally and was able to add the DOI that caused the issue:

image


gzurowski:
We are deprecating this repository in favor of the new application repository (https://github.com/ResearchHub/web).

PR #1773, opened 2024-07-13
Add Persona verification step to Verification v2

gzurowski:
- Add dependencies for the Persona React component.
- Add new step form for verifying with Persona.
- Integrate with Verification v2 modal.

image


yattias (approved):
LGTM

PR #1783, opened 2024-07-27
[Identity verification] Listen to webhook to know if verification is successful (instead of persona default)

yattias:
- Listen to webhook to know if verification is successful
- Edit profile modal

image


koutst (approved):

gzurowski (approved):
LGTM 👍

PR #1784, opened 2024-07-27
Set "Author" as the title for the author profile page

gzurowski:
Set a title for the author profile page instead of the default _Explore ResearchHub_.

### DIFF
--- a/components/Navbar.js
+++ b/components/Navbar.js
@@ -60,6 +60,7 @@ const Navbar = (props) => {
   const [stickyNavbar, setStickyNavbar] = useState(true);
   const { open, close } = useWeb3Modal();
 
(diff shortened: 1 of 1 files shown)


yattias (commented):
This looks good. There is a version of the back button that will take user to prev page they were on. Not in front of computer atm - if this one does so then great. If not, check out document page

gzurowski:
> This looks good. There is a version of the back button that will take user to prev page they were on. Not in front of computer atm - if this one does so then great. If not, check out document page

This change is using the existing `BackBtn` component that has some history logic for navigating back, but it doesn't seem to work properly. As discussed, proceeding to merge the change and we'll return to this later.

yattias (approved):

PR #1819, opened 2024-08-26
Use new achievements field

koutst:


yattias (commented):
LGTM

yattias (approved):

components/Author/lib/types.ts
```diff
@@ -110,7 +109,16 @@ function ensureSufficientYears(activityData: YearlyActivity[]): YearlyActivity[]
 }  
 
 export const parseAuthorAchievements = (raw: any): Array<Achievement> => {
-  return raw.achievements || [];
+  let achievements: Achievement[] = [];
+  for (const key in raw.achievements) {
+    if (raw.achievements.hasOwnProperty(key)) {
+      const value = raw.achievements[key];
```
yattias, line 115: Can we update a bit for readability to something like:

```
const achievement = raw.achievement[key];
if (achievement["value"] >= value["milestones"][0]) ...
```

PR #1855, opened 2024-10-02
Display hubs used for rep first

yattias:


components/Author/Tabs/FeedCard.tsx
```diff
@@ -64,7 +64,7 @@ export type FeedCardProps = {
   created_by: any;
   document: any;
   created_date: any;
-  documentFilter: any;
+  documentFilterdocumentFilter: any;
```
gzurowski, line 67: Was this change intended?

components/Author/Tabs/FeedCard.tsx
```diff
@@ -165,7 +165,12 @@ function FeedCard({
 
   const visibleDate = parsedDoc?.publishedDate || parsedDoc?.createdDate;
 
-  const parsedHubs = (hubs || []).map(parseHub);
+  // Sort hubs by displaying ones with isUsedForRep first.
+  // This approach should be deprecated once we introduce "hub" namespace.
```
gzurowski, line 169:
```suggestion
  // TODO: This approach should be deprecated once we introduce "hub" namespace.
```

gzurowski (approved):
LGTM 👍

components/Author/Tabs/FeedCard.tsx
yattias: Good catch @gzurowski probably some keyboard error

PR #1858, opened 2024-10-12
Remove empty tracking event calls

gzurowski:
Remove all tracking events targeting `events/amplitude/forward_event` which is a no-op view endpoint on the backend (see https://github.com/ResearchHub/researchhub-backend/blob/81a77791b8c60066274f3a72e64f47698b444deb/src/analytics/views.py#L107).


yattias (approved):

yattias:
@gzurowski looks good. Can you just verify creating new account works properly before merging?

PR #1862, opened 2024-10-18
fix: Correctly display "member for" period in years, months, days

gzurowski:
Take years and months already displayed into account.

So instead of:

> Member since 2 years, 27 months and 851 days

Show this:

> Member since 2 years, 4 months and 1 day

Fixes https://github.com/ResearchHub/issues/issues/122


gzurowski:
Successfully tested:

image

Instead of:

image

yattias (approved):

PR #1874, opened 2024-11-02
Fix author dates

TylerDiorio:
Fixing the dates shown on the author profiles to be the correct amount of days.
- Change config/utils/dates.js 'specificTimeSince(date)' which is only used in the components/Author/Profile/AuthorProfileHeader.tsx file so there shouldn't be any off-target effects

See Issue #137 https://github.com/ResearchHub/issues/issues/137


TylerDiorio:
Before:
Screenshot 2024-11-02 at 1 08 01 PM

After:
Screenshot 2024-11-02 at 1 08 06 PM

PR #1918, opened 2025-04-27
Render TipTap Mention Blocks in Comments

nicktytarenko:
## What?
- This PR introduces support for rendering TipTap mention blocks within the comments.

## How?
- **User/Author** Mentions: Render as clickable `@username` links that navigate to the user’s profile.
- **Paper/Post** Mentions: Render as links to the referenced paper or post.
    > **Note**: For paper and post mentions, links will redirect to the new ResearchHub app. This is because the mention data does not include the document slug, which is required for routing in the current (old) app. As a result, we cannot generate valid old-app URLs for these entities.

image


gzurowski:
We are deprecating this repository in favor of the new application repository (https://github.com/ResearchHub/web).

